PR #2, opened 2016-06-21
ENH: At least one test per function

ahgnaw:
Made some functions more general and added at least one test per function.

### DIFF
--- a/qrisk/stats.py
+++ b/qrisk/stats.py
@@ -15,13 +15,10 @@
 
 from __future__ import division
 
-from collections import OrderedDict
-from functools import partial
-
 import pandas as pd
 import numpy as np
 import scipy as sp
-import scipy.stats
+import scipy.stats  # noqa
 
 
 APPROX_BDAYS_PER_MONTH = 21
@@ -41,17 +38,23 @@ ANNUALIZATION_FACTORS = {
     MONTHLY: MONTHS_PER_YEAR
 }
 
-def cum_returns(returns, starting_value=None):
+
+def cum_returns(returns, starting_value=0):
     """
     Compute cumulative returns from simple returns.
 
     Parameters
     ----------
     returns : pd.Series
-        Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        Returns of the strategy as a percentage, noncumulative.
+         - Time series with decimal returns.
+         - Example:
+            2015-07-16    -0.012143
+            2015-07-17    0.045350
+            2015-07-20    0.030957
+            2015-07-21    0.004902.
     starting_value : float, optional
-       The starting returns (default 1).
+       The starting returns (default is 0).
 
     Returns
     -------
@@ -62,10 +65,12 @@ def cum_returns(returns, starting_value=None):
     -----
     For increased numerical accuracy, convert input to log returns
     where it is possible to sum instead of multiplying.
+    PI((1+r_i)) - 1 = exp(ln(PI(1+r_i)))     # x = exp(ln(x))
+                    = exp(SIGMA(ln(1_r_i))   # ln(a*b) = ln(a) + ln(b)
     """
 
     # df_price.pct_change() adds a nan in first position, we can use
-    # that to have cum_returns start at the origin so that
+    # that to have cum_logarithmic_returns start at the origin so that
     # df_cum.iloc[0] == starting_value
     # Note that we can't add that ourselves as we don't know which dt
     # to use.
@@ -74,21 +79,21 @@ def cum_returns(returns, starting_value=None):
 
     df_cum = np.exp(np.log(1 + returns).cumsum())
 
-    if starting_value is None:
+    if starting_value is 0:
         return df_cum - 1
     else:
         return df_cum * starting_value
 
 
-def aggregate_returns(df_daily_rets, convert_to):
+def aggregate_returns(returns, convert_to):
     """
     Aggregates returns by week, month, or year.
 
     Parameters
     ----------
-    df_daily_rets : pd.Series
+    returns : pd.Series
        Daily returns of the strategy, noncumulative.
-        - See full explanation in tears.create_full_tear_sheet (returns).
+        - See full explanation in cum_returns.
     convert_to : str
         Can be 'weekly', 'monthly', or 'yearly'.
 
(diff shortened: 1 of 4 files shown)


qrisk/stats.py
```diff
@@ -41,17 +40,23 @@
...
+def cum_returns(returns, starting_value=0):
     """
     Compute cumulative returns from simple returns.
 
     Parameters
     ----------
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+         - Time series with decimal returns.
```
abhijeetkalyan: With the `returns` metric, do you think it's worth clarifying whether they are % returns or absolute dollar returns? (From the examples, I feel like it's reasonably clear that they are percentages, but wondering if it's worth clarifying anyways)

qrisk/stats.py
```diff
@@ -62,33 +67,35 @@ def cum_returns(returns, starting_value=None):
...
     # df_price.pct_change() adds a nan in first position, we can use
-    # that to have cum_returns start at the origin so that
+    # that to have cum_logarithmic_returns start at the origin so that
     # df_cum.iloc[0] == starting_value
     # Note that we can't add that ourselves as we don't know which dt
     # to use.
-    if pd.isnull(returns.iloc[0]):
-        returns.iloc[0] = 0.
+    if pd.isnull(returns[0]):
+        returns[0] = 0.
```
twiecki: Why this change? It makes the index more ambiguous.

qrisk/stats.py
```diff
@@ -166,67 +149,81 @@ def max_drawdown(returns):
...
     cum_max_return = df_cum_rets.cummax()
+    drawdown = df_cum_rets.sub(cum_max_return).div(cum_max_return)
 
-    return df_cum_rets.sub(cum_max_return).div(cum_max_return).min()
+    return drawdown.min()
 
 
-def annual_return(returns, period=DAILY):
-    """Determines the annual returns of a strategy.
+def annual_return(returns, period=DAILY, annualization=ANNUALIZATION_FACTORS):
```
twiecki: Any reason why this should be API?

qrisk/stats.py
```diff
@@ -238,37 +235,43 @@ def annual_volatility(returns, period=DAILY):
...
-            " Must be '{}', '{}', or '{}'".format(
-                period, DAILY, WEEKLY, MONTHLY
+            "period cannot be '{}'. "
+            "Can be '{}'.".format(
+                period, "', '".join(annualization.keys())
             )
         )
 
-    return returns.std() * np.sqrt(ann_factor)
+    volatility = returns.std() * pow(ann_factor, float(1.0/alpha))
```
twiecki: Why this change? `sqrt` seems more readable.

qrisk/stats.py
```diff
@@ -328,21 +335,77 @@ def omega_ratio(returns, annual_return_threshhold=0.0):
         return np.nan
 
 
-def sortino_ratio(returns, required_return=0, period=DAILY):
+def ex_ante_sharpe_ratio(returns, risk_free=0, period=DAILY,
```
twiecki: is there an `ex_post` sharpe ratio?

qrisk/stats.py
```diff
@@ -499,57 +539,64 @@ def alpha_beta(returns, factor_returns):
...
         Alpha.
-"""
+    """
 
-    return alpha_beta(returns, factor_returns)[0]
+    return alpha_beta(returns, benchmark, risk_free=risk_free)[0]
 
 
-def beta(returns, factor_returns):
+def beta(returns, benchmark, risk_free=0.0):
```
twiecki: We pretty consistently use `factor_returns` in pyfolio.

twiecki:
Before merging, we should make sure that pyfolio is still compatible, otherwise getting them in sync will be much harder.

qrisk/stats.py
ahgnaw: Futures have different number of business days in a year than equities. Sometimes it is dependent on the commodity at hand. This is part of the API so that qrisk is futures compatible.

qrisk/stats.py
ahgnaw: Good point, i'll change it back

qrisk/stats.py
ahgnaw: Not yet, this first pr has the most commonly used sharpe ratio.

qrisk/tests/test_stats.py
```diff
@@ -40,12 +40,71 @@ class TestStats(TestCase):
...
+    ])
+    def test_aggregate_returns(self, returns, convert_to, expected):
+        self.assertEqual(
+            len(stats.aggregate_returns(returns, convert_to)),
+            expected)
+
+    @parameterized.expand([
+        (simple_rets, 0.0),
+        (s_3, -0.10000000000000002)
+        ])
```
abhijeetkalyan: This is really pedantic, but on some tests we do:
```python
@parameterized.expand([
    foo,
    bar
    ])
```

whereas on others, we do
```python
@parameterized.expand([
    foo,
    bar
])
```

can we be consistent on one style? (Not a big deal though)

abhijeetkalyan:
If we're getting `travis` set up as a part of this PR, it might be worth adding a little more scope and setting up `coverage` as well, so we can get an idea of how much of our code is tested.

qrisk/stats.py
ahgnaw: When calculating volatility, it is sometimes useful to use slightly different values here. There might be some users that want the flexibility.

https://en.wikipedia.org/wiki/Volatility_(finance)#Mathematical_definition

qrisk/stats.py
ahgnaw: Will change it.

qrisk/stats.py
ahgnaw: Yes, I think it's worth clarifying.

ahgnaw:
@twiecki the one incompatibility is the change from 'sharpe_ratio' to 'ex_ante_sharpe_ratio', it should be a straightforward change once qrisk is used in pyfolio.

ahgnaw:
@abhijeetkalyan Adding CI will be its own pr. `coverage` can definitely be included.

qrisk/stats.py
twiecki: I would just stick with `sharpe_ratio` then, principle of least surprise.

qrisk/stats.py
```diff
@@ -238,37 +235,43 @@ def annual_volatility(returns, period=DAILY):
...
+        Factor used to convert returns into annual returns.
+        - See full explanation in annual_return.
 
 
     Returns
     -------
-    float
-        Calmar ratio (drawdown ratio).
+    float, np.nan
+        Calmar ratio (drawdown ratio) as float. Returns np.nan is there is no
```
twiecki: typo is -> if

qrisk/stats.py
```diff
@@ -290,18 +294,22 @@ def calmar_ratio(returns, period=DAILY):
...
     ----------
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
-    annual_return_threshold : float, optional
-        Threshold over which to consider positive vs negative
-        returns. For the ratio, it will be converted to a daily return
-        and compared to returns.
+        - See full explanation in cum_returns.
+    mar : float, optional
```
twiecki: I would just call it `minimum_acceptance_return`, more descriptive always beats saving a few characters.

qrisk/stats.py
```diff
@@ -472,25 +508,30 @@ def information_ratio(returns, factor_returns):
     See https://en.wikipedia.org/wiki/information_ratio for more details.
 
     """
+
     active_return = returns - factor_returns
-    tracking_error = np.std(active_return, ddof=1)
```
twiecki: I would stick with the `np` prefix, makes it clear to everyone reading the code which one is being used.

PR #3, opened 2016-06-27
ENH: Added Travis CI

ahgnaw:
Adapted nosetests to pass with Travis. Added travis build file.


richafrank:
@llllllllll Would you also take a look at the CI install here?

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+    fi
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda  
+
+install:
+  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION numpy scipy nose pandas flake8
+  #- pip install --no-deps git+https://github.com/quantopian/zipline
```
llllllllll: cruft?

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
+  - source activate testenv
+  - pip install nose_parameterized
```
llllllllll: we should just be using something like `pip install -e .[dev]`. Right now it looks like the requirements need to be maintaned here and the setup.py.

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+
+before_script:
+  - "flake8 ."
+
+script:
+  - nosetests
+  - source deactivate
+
+notifications:
+  email: false
```
llllllllll: newline

abhijeetkalyan:
@llllllllll got this

.travis.yml
ahgnaw: I followed [this](http://conda.pydata.org/docs/travis.html) when setting up miniconda. Conda is used to successfully install scipy and numpy in travis.

llllllllll: I was commenting on the commented out zipline install

llllllllll:
Why did all these changes leak into the "add travis" pr?

qrisk/stats.py
```diff
@@ -41,17 +38,23 @@
...
+        Returns of the strategy as a percentage, noncumulative.
+         - Time series with decimal returns.
+         - Example:
+            2015-07-16    -0.012143
+            2015-07-17    0.045350
+            2015-07-20    0.030957
+            2015-07-21    0.004902.
     starting_value : float, optional
-       The starting returns (default 1).
+       The starting returns (default is 0).
```
llllllllll: We don't need to specify the default in the docs if the default value is present in the signature

qrisk/stats.py
```diff
@@ -74,21 +79,21 @@ def cum_returns(returns, starting_value=None):
 
     df_cum = np.exp(np.log(1 + returns).cumsum())
 
-    if starting_value is None:
+    if starting_value is 0:
```
llllllllll: we should use `==` to compare ints

qrisk/stats.py
```diff
@@ -74,21 +79,21 @@ def cum_returns(returns, starting_value=None):
...
 
 
-def aggregate_returns(df_daily_rets, convert_to):
+def aggregate_returns(returns, convert_to):
     """
     Aggregates returns by week, month, or year.
 
     Parameters
     ----------
-    df_daily_rets : pd.Series
```
llllllllll: did this used to be called `df_faily_rets` when it was a series?

qrisk/stats.py
```diff
@@ -472,25 +510,32 @@ def information_ratio(returns, factor_returns):
...
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in cum_returns.
     factor_returns : pd.Series
          Daily noncumulative returns of the factor to which beta is
          computed. Usually a benchmark such as the market.
          - This is in the same style as returns.
+    risk_free : int, float, optional
+        Constant risk-free return throughout the period. For example, the
+        interest rate on a three month us treasury bill.
+        - Default is 0.
```
llllllllll: no need for default when it is in the signature

setup.py
```diff
@@ -5,9 +5,11 @@
...
 
 DISTNAME = 'qrisk'
-DESCRIPTION = "qrisk is a Python library with performance and risk statistics commonly used in quantitative finance"
-LONG_DESCRIPTION = """qrisk is a Python library with performance and risk statistics
-commonly used in quantitative finance by `Quantopian Inc`_. It is used by `Zipline`_ open source backtesting library as well as `pyfolio`_.
+DESCRIPTION = """qrisk is a Python library with performance and risk statistics
+commonly used in quantitative finance"""
+LONG_DESCRIPTION = """qrisk is a Python library with performance and risk
+statistics commonly used in quantitative finance by `Quantopian Inc`_. It is
+used by `Zipline`_ open source backtesting library as well as `pyfolio`_.
```
llllllllll: It isn't used byt zipline yet

qrisk/stats.py
```diff
@@ -499,57 +544,72 @@ def alpha_beta(returns, factor_returns):
...
-    ret_index = returns.index
-    beta, alpha = sp.sp.stats.linregress(factor_returns.loc[ret_index].values,
-                                      returns.values)[:2]
+    y = (returns - risk_free).loc[factor_returns.index].dropna()
+    x = (factor_returns - risk_free).loc[y.index].dropna()
+    y = y.loc[x.index]
+    beta, alpha = sp.stats.linregress(x.values, y.values)[:2]
 
-    return alpha * APPROX_BDAYS_PER_YEAR, beta
+    return alpha, beta
```
llllllllll: why are we no longer multiplying alpha by 252?

qrisk/stats.py
```diff
@@ -328,21 +334,78 @@ def omega_ratio(returns, annual_return_threshhold=0.0):
...
+        raise ValueError(
+            "period cannot be '{}'. "
+            "Can be '{}'.".format(
+                period, "', '".join(annualization.keys())
+            )
+        )
+
+    returns_risk_adj = returns - risk_free
+
+    if (len(returns_risk_adj) < 5) or np.all(returns_risk_adj == 0):
```
llllllllll: why is this less than 5?

.travis.yml
```diff
@@ -0,0 +1,38 @@
...
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    fi
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda  
+
+install:
+  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION numpy scipy
```
llllllllll: we should pin the numpy / scipy versions. Right now we always test on the latest.

qrisk/__init__.py
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .stats import *
+from .stats import *  # noqa
```
llllllllll: can we just kill the import *?

qrisk/stats.py
```diff
@@ -15,13 +15,10 @@
...
 from __future__ import division
 
-from collections import OrderedDict
-from functools import partial
-
 import pandas as pd
 import numpy as np
 import scipy as sp
-import scipy.stats
+import scipy.stats  # noqa
```
llllllllll: we should just remove the import if it is not used

qrisk/stats.py
```diff
@@ -149,7 +130,7 @@ def max_drawdown(returns):
     ----------
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in cum_returns.
```
llllllllll: can we make these sphinx links like:
```
:func:`~path.to.cum_returns`
```

qrisk/stats.py
```diff
@@ -166,67 +147,81 @@ def max_drawdown(returns):
...
 
     num_years = float(len(returns)) / ann_factor
-    df_cum_rets = cum_returns(returns, starting_value=100)
     start_value = 100
-    end_value = df_cum_rets.iloc[-1]
-
+    end_value = cum_returns(returns, starting_value=start_value).iloc[-1]
     total_return = (end_value - start_value) / start_value
-    annual_return = (1. + total_return) ** (1 / num_years) - 1
+    annual_return = pow((1. + total_return), (1 / num_years)) - 1
```
llllllllll: why did we swap pow for `**`?

qrisk/stats.py
```diff
@@ -238,37 +233,43 @@ def annual_volatility(returns, period=DAILY):
...
-            " Must be '{}', '{}', or '{}'".format(
-                period, DAILY, WEEKLY, MONTHLY
+            "period cannot be '{}'. "
+            "Can be '{}'.".format(
+                period, "', '".join(annualization.keys())
             )
         )
 
-    return returns.std() * np.sqrt(ann_factor)
+    volatility = returns.std() * pow(ann_factor, float(1.0/alpha))
```
llllllllll: Can we put a space between binary operators?

.gitignore
```diff
@@ -60,3 +60,6 @@ target/
 
 #Ipython Notebook
 .ipynb_checkpoints
+
+#PyPI Configuration
+.pypirc
```
llllllllll: This file should be in your home directory, does it read from the local dir also?

qrisk/stats.py
```diff
@@ -102,45 +107,21 @@ def cumulate_returns(x):
         return cum_returns(x)[-1]
 
     if convert_to == WEEKLY:
-        return df_daily_rets.groupby(
+        return returns.groupby(
```
llllllllll: note: out of scope for this pr, but if this only supports pandas 18 we can replace these groupbys with resamples which will be much faster

qrisk/stats.py
```diff
@@ -166,67 +147,81 @@ def max_drawdown(returns):
...
     ----------
     returns : pd.Series
         Periodic returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in cum_returns.
     period : str, optional
         - defines the periodicity of the 'returns' data for purposes of
         annualizing. Can be 'monthly', 'weekly', or 'daily'
         - defaults to 'daily'.
+    annualization : dict, optional
```
llllllllll: The api here seems unintuitive. to pass a custom period, I would need to pass both the key and the dict so that the function call getitem for me. How do you feel about just passing the numeric value?

qrisk/stats.py
```diff
@@ -238,37 +233,43 @@ def annual_volatility(returns, period=DAILY):
...
+        Factor used to convert returns into annual returns.
+        - See full explanation in annual_return.
 
 
     Returns
     -------
-    float
-        Calmar ratio (drawdown ratio).
+    float, np.nan
```
llllllllll: np.nan is a float

qrisk/tests/test_stats.py
```diff
@@ -40,32 +40,90 @@ class TestStats(TestCase):
...
+    d_week = {'one': pd.Series(one, index=df_index_week),
+              'two': pd.Series(two, index=df_index_week)}
+    df_week = pd.DataFrame(d_week)
+
+    d_month = {'one': pd.Series(one, index=df_index_month),
+               'two': pd.Series(two, index=df_index_month)}
+    df_month = pd.DataFrame(d_month)
+
+    @parameterized.expand([
+        (simple_rets, 0, 0.33100000000000041),
```
llllllllll: since we are doing almost equal should we remove the insignificant digits in these expecteations. They seem to add noise to the expected value.

.travis.yml
```diff
@@ -0,0 +1,38 @@
...
+  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
```
llllllllll: I don't think we use enum34

qrisk/tests/test_stats.py
```diff
@@ -40,32 +40,90 @@ class TestStats(TestCase):
...
+        (simple_week_rets, 100, 133.10000000000005),
+        (simple_benchmark, 100, 112.55088100000002)
+    ])
+    def test_cum_returns(self, returns, starting_value, expected):
+        self.assertAlmostEqual(
+            stats.cum_returns(
+                returns, starting_value
+            )[-1],
```
llllllllll: should we check the whole array?

qrisk/tests/test_stats.py
llllllllll: do we want a test case that starts with `nan` to make sure we hit that branch in `cum_returns`?

qrisk/tests/test_stats.py
```diff
@@ -40,32 +40,90 @@ class TestStats(TestCase):
...
+            expected,
+            DECIMAL_PLACES)
+
+    @parameterized.expand([
+        (simple_rets, stats.WEEKLY, 72),
```
llllllllll: same here, we should test the start of the series as `nan`

qrisk/tests/test_stats.py
```diff
@@ -40,32 +40,90 @@ class TestStats(TestCase):
...
+        (simple_month_rets, stats.YEARLY, 42)
+    ])
+    def test_aggregate_returns(self, returns, convert_to, expected):
+        self.assertEqual(
+            len(stats.aggregate_returns(returns, convert_to)),
+            expected)
+
+    @parameterized.expand([
+        (simple_rets, 0.0),
+        (s_3, -0.10000000000000002)
```
llllllllll: why is this -0.10 instead of -0.20 if we went from +10 to -10?

qrisk/tests/test_stats.py
```diff
@@ -40,32 +40,90 @@ class TestStats(TestCase):
...
                 returns,
                 period=period
             ),
-            expected)
+            expected,
+            DECIMAL_PLACES)
 
     @parameterized.expand([
         (simple_rets, stats.DAILY, 0.12271674212427248),
-        (simple_rets, stats.DAILY, 0.12271674212427248),
```
llllllllll: why are we removing these test cases?

qrisk/tests/test_stats.py
```diff
@@ -79,49 +137,140 @@ def test_annual_volatility(self, returns, period, expected):
         )
 
     @parameterized.expand([
-        (simple_rets, 1.2333396776895436),
+        (s_2.pct_change().dropna(), -2.3992211554712197)
```
llllllllll: can we craft an input that gives a more intuitive almar ratio? I have no intuition for how accurate this is.

qrisk/stats.py
```diff
@@ -74,21 +79,21 @@ def cum_returns(returns, starting_value=None):
 
     df_cum = np.exp(np.log(1 + returns).cumsum())
```
llllllllll: `log1p`

qrisk/tests/test_stats.py
```diff
@@ -79,49 +137,140 @@ def test_annual_volatility(self, returns, period, expected):
...
+    ])
+    def test_calmar(self, returns, expected):
+        self.assertAlmostEqual(
+            stats.calmar_ratio(
+                returns),
+            expected,
+            DECIMAL_PLACES)
+
+    @parameterized.expand([
+        (s_1, 0.0, 2.0)
```
llllllllll: should we add some cases with a non-zerio minimum threshold, this is just positive returns over negative returns right now.

qrisk/tests/test_stats.py
```diff
@@ -79,49 +137,140 @@ def test_annual_volatility(self, returns, period, expected):
...
+
+    @parameterized.expand([
+        (simple_rets, 1.2321057207245873),
         (np.zeros(10), np.nan),
         ([0.1, 0.2, 0.3], np.nan)
     ])
-    def test_sharpe(self, returns, expected):
+    def test_sharpe_ratio(self, returns, expected):
         assert_almost_equal(
             stats.sharpe_ratio(
```
llllllllll: we should test with a non-zero risk free rate.

qrisk/tests/test_stats.py
```diff
@@ -79,49 +137,140 @@ def test_annual_volatility(self, returns, period, expected):
...
     @parameterized.expand([
-        (pd.Series(px_list_2,
-                   index=dt_2).pct_change().dropna(), -2.3992211554712197)
+        (simple_rets, 0.0, stats.DAILY, 0.0),
+        (simple_week_rets, 0.0, stats.WEEKLY, 0.0),
+        (simple_month_rets, 0.0, stats.MONTHLY, 0.0),
+        (simple_rets, 0.1, stats.DAILY, 1.5826812692390093),
+        (simple_week_rets, 0.1, stats.WEEKLY, 0.718943669559723),
+        (simple_month_rets, 0.1, stats.MONTHLY, 0.3453693674893592),
+        (df_simple, 0.0, stats.DAILY,
```
llllllllll: the order of the input columns is undefined, will this cause a problem comparing the results?

llllllllll: also, how did we come up with the inputs to get these expected outputs? I am trying to understand how I can know this is correct based on these results.

qrisk/tests/test_stats.py
```diff
@@ -79,49 +137,140 @@ def test_annual_volatility(self, returns, period, expected):
...
+        (s_3, 0.49842309053331446)
+    ])
+    def test_stability_of_timeseries(self, returns, expected):
+        self.assertAlmostEqual(
+            stats.stability_of_timeseries(returns),
+            expected,
+            DECIMAL_PLACES)
+
+    @parameterized.expand([
+        (np.random.randn(10000), 1.)
```
llllllllll: should we test a non-normal distrobution?

setup.py
```diff
@@ -17,32 +19,45 @@
 MAINTAINER_EMAIL = 'opensource@quantopian.com'
 AUTHOR = 'Quantopian Inc'
 AUTHOR_EMAIL = 'opensource@quantopian.com'
-URL = "https://github.com/quantopian/pyfolio"
+URL = "https://github.com/quantopian/qrisk"
```
llllllllll: can we leave the quotes consistent?

qrisk/tests/test_stats.py
```diff
@@ -79,49 +137,140 @@ def test_annual_volatility(self, returns, period, expected):
...
+            expected,
+            DECIMAL_PLACES)
 
-    def test_tail_ratio(self):
-        returns = np.random.randn(10000)
+    @parameterized.expand([
+        (simple_rets, simple_benchmark, (0.00020161290322580659,
+                                         3.3266129032258061))
+    ])
+    def test_alpha_beta(self, returns, benchmark, expected):
```
llllllllll: maybe we should test `alpha` and `beta` with `alpha_beta` to ensure that the results are the same.

.gitignore
ahgnaw: No it doesn't I'll remove it.

qrisk/stats.py
ahgnaw: Only the  name changed, I prefer 'returns' over 'df_daily_rets'.

qrisk/stats.py
ahgnaw: Sure, flake8 didn't mention it.

qrisk/stats.py
ahgnaw: gp

qrisk/stats.py
ahgnaw: It might be arbitrary. @twiecki?

qrisk/stats.py
ahgnaw: Positively - I'll change it.

qrisk/stats.py
ahgnaw: It's not part of the typical alpha calculation. There is such a thing as annualized alpha but it's not the basic definition.

qrisk/tests/test_stats.py
ahgnaw: They are duplicates. Is there a reason to test the same case twice?

llllllllll: ah, yeah we should remove the duplicates, sorry.

setup.py
ahgnaw: Will do.

qrisk/tests/test_stats.py
ahgnaw: We should and we will.

qrisk/tests/test_stats.py
ahgnaw: Ok.

setup.py
ahgnaw: GP

qrisk/tests/test_stats.py
ahgnaw: The first three test when downside risk should be 0 at different time periods. The fourth through sixth test the same inputs but with a higher minimum acceptable return, again for different time periods. The last three test when the input is a dataframe and I believe the values for 'one' and 'two are a list of random  numbers with a mean of 0 and 0.15.

qrisk/tests/test_stats.py
twiecki: Probably better to add a test with a shifted normal (e.g. mu = .1).

qrisk/stats.py
twiecki: Mostly arbitrary, yes. If anything should be higher.

qrisk/stats.py
twiecki: +1

qrisk/stats.py
ahgnaw: I can't explain why, but if we remove this line then sp.stats* functions won't work.

twiecki: `scipy` does not import `stats` by default so it's required. flake8 is just not smart enough to track `scipy.stats` to `sp.stats`.

llllllllll: I would argue that because you were confused by this, other readers will be to. We can just replace it with something like `import scipy.stats as sps` or `from scipy.stats import {}`

twiecki: Never seen `{}`. I like the first, or just do `from scipy import stats`.

llllllllll: the `{}` was like a format string, or `%s`, it meant "put the name of the function(s) you want to use here

qrisk/tests/test_stats.py
ahgnaw: Excellent catch.

qrisk/stats.py
ahgnaw: I like `from scipy import stats`, will use that one

qrisk/__init__.py
ahgnaw: Yes.

.travis.yml
```diff
@@ -0,0 +1,43 @@
+language: python
+sudo: false
+
+python:
+  - "2.7"
+  - "3.4"
```
twiecki: 3.5

qrisk/stats.py
twiecki: +1

qrisk/stats.py
```diff
@@ -164,69 +140,84 @@ def max_drawdown(returns):
...
-        - defines the periodicity of the 'returns' data for purposes of
-        annualizing. Can be 'monthly', 'weekly', or 'daily'
-        - defaults to 'daily'.
+        Defines the periodicity of the 'returns' data for purposes of
+        annualizing. Can be 'monthly', 'weekly', or 'daily'.
+    annualization : int, optional
+        Factor used to convert the returns into annual returns, if different
+        from the default values used for different periods.
+        - Default:
+          annualization = {'daily': 252,
```
twiecki: Needs updating.

qrisk/stats.py
```diff
@@ -353,15 +431,22 @@ def sortino_ratio(returns, required_return=0, period=DAILY):
...
-    except KeyError:
-        raise ValueError(
-            "period cannot be: '{}'."
-            " Must be '{}', '{}', or '{}'".format(
-                period, DAILY, WEEKLY, MONTHLY
+
+    if len(returns) < 2:
+        return np.nan
+
+    if annualization is None:
```
twiecki: Does it make sense to turn that into a function?

qrisk/stats.py
```diff
@@ -609,8 +699,6 @@ def tail_ratio(returns):
     max_drawdown,
     omega_ratio,
     sortino_ratio,
-    sp.stats.skew,
```
twiecki: I think we'll want to keep those.

setup.py
```diff
@@ -4,45 +4,59 @@
...
-               'Topic :: Scientific/Engineering',
-               'Topic :: Scientific/Engineering :: Mathematics',
-               'Operating System :: OS Independent']
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 2",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 2.7",
+    "Programming Language :: Python :: 3.4",
```
twiecki: 3.5

twiecki:
See comments, afterwards good to merge.

qrisk/stats.py
ahgnaw: There are some years that don't have 252 business days in them. 2017 is one of them.

.travis.yml
```diff
@@ -0,0 +1,53 @@
...
+    fi
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda  
+
+install:
+  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION scipy=$SCIPY_VERSION libgfortran=$LIBGFORTRAN_VERSION
+  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
```
llllllllll: Are we using enum34 anywhere?

.travis.yml
```diff
@@ -0,0 +1,53 @@
...
+  - source activate testenv
+  - pip install -e .[dev]
+  - python setup.py build_ext --inplace
```
llllllllll: this will be done by the line above it

README.md
```diff
@@ -1,2 +1,11 @@
 # qrisk
 Common financial risk metrics.
+
+## Installation
+
+### Installing with `pip`
+
+```
+pip install git+git://github.com/quantopian/qrisk.git
```
llllllllll: isn't this on pypi now, shouldn't it be `pip install qrisk`?

qrisk/__init__.py
```diff
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .stats import *
-
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+from .stats import *  # noqa
```
llllllllll: Can we be explicit about what we are importing?

qrisk/stats.py
```diff
@@ -62,33 +64,35 @@ def cum_returns(returns, starting_value=None):
...
     # to use.
     if pd.isnull(returns.iloc[0]):
         returns.iloc[0] = 0.
 
-    df_cum = np.exp(np.log(1 + returns).cumsum())
+    df_cum = np.exp(np.log1p(returns).cumsum())
 
-    if starting_value is None:
+    if starting_value == 0:
         return df_cum - 1
```
llllllllll: The docstring shows that the default has changed from 1 to 0, do we need to change this line?

llllllllll: also, since the default is now an int, can we just define the formula without the branch?

qrisk/stats.py
```diff
@@ -62,33 +64,35 @@ def cum_returns(returns, starting_value=None):
...
     """
     Aggregates returns by week, month, or year.
 
     Parameters
     ----------
-    df_daily_rets : pd.Series
+    returns : pd.Series
        Daily returns of the strategy, noncumulative.
-        - See full explanation in tears.create_full_tear_sheet (returns).
+        - See full explanation in :func:`cum_returns`.
```
llllllllll: does this get linked to correctly? I think you may need to pass the full path, like `pyfolio.something.cum_returns`. In sphinx, you can prefix the path with `~` to have it only show the name of the function though, like:
```
:func:`~a.b.c`
```
will render as `c`

qrisk/stats.py
```diff
@@ -164,69 +140,84 @@ def max_drawdown(returns):
     if returns.size < 1:
         return np.nan
 
-    df_cum_rets = cum_returns(returns, starting_value=100)
-    cum_max_return = df_cum_rets.cummax()
+    current_max = returns.copy()
+    for i in range(1, current_max.size):
```
llllllllll: why did we replace cummax with the looping version?

ahgnaw: This was resulting in the wrong answer so I changed it to the version I know works.

llllllllll: we changed the formula other than the cummax though, I don't think it was the cummax that was the problem. This look will be very slow for large inputs.

qrisk/stats.py
```diff
@@ -237,50 +228,60 @@ def annual_volatility(returns, period=DAILY):
...
+                "Period cannot be '{}'. "
+                "Can be '{}'.".format(
+                    period, "', '".join(ANNUALIZATION_FACTORS.keys())
+                )
             )
-        )
+    else:
+        ann_factor = annualization
+
+    volatility = returns.std() * (ann_factor ** float(1.0 / alpha))
```
llllllllll: why are we calling float on the result of the div?

qrisk/stats.py
```diff
@@ -314,10 +323,18 @@ def omega_ratio(returns, annual_return_threshhold=0.0):
 
     """
 
-    daily_return_thresh = pow(1 + annual_return_threshhold, 1 /
-                              APPROX_BDAYS_PER_YEAR) - 1
+    if returns.size < 2:
```
llllllllll: `len(returns)`

qrisk/stats.py
```diff
@@ -314,10 +323,18 @@ def omega_ratio(returns, annual_return_threshhold=0.0):
...
+        return np.nan
+
+    if annualization == 1:
+        return_threshold = required_return
+    elif required_return <= -1:
+        return np.nan
+    else:
+        return_threshold = (1 + required_return) ** \
+            float(1 / annualization) - 1
```
llllllllll: do we want the truncated division in py2 here?

qrisk/stats.py
llllllllll: it is just an approximation, there are other years that have less trading days, for example when hurricane sandy closed the market.

llllllllll: also, I think that Thomas was suggesting that we pull this try/except out into a function since many functions have the same duplicated code.

qrisk/stats.py
```diff
@@ -499,57 +569,72 @@ def alpha_beta(returns, factor_returns):
     float
         Beta.
 
-"""
+    """
+    if len(returns) < 2:
+        return (np.nan, np.nan)
```
llllllllll: parens are not needed here (only commented since it is inconsistent with the return below)

qrisk/stats.py
```diff
@@ -499,57 +569,72 @@ def alpha_beta(returns, factor_returns):
...
     float
         Beta.
-"""
+    """
 
-    return alpha_beta(returns, factor_returns)[1]
+    if len(returns) < 2:
+        return np.nan
+
+    return alpha_beta(returns, factor_returns, risk_free=risk_free)[1]
```
llllllllll: Can we make this return a namedtuple so that it can be `alpha_beta(...).beta`?

qrisk/stats.py
```diff
@@ -561,17 +646,19 @@ def stability_of_timeseries(returns):
     ----------
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`cum_returns`.
 
     Returns
     -------
     float
         R-squared.
```
llllllllll: this says r^2 but the code returns rhat

qrisk/tests/test_stats.py
```diff
@@ -79,49 +158,230 @@ def test_annual_volatility(self, returns, period, expected):
...
+                        required_return=required_return,
+                        period=period)
+        if isinstance(downside_risk, float):
+            assert_almost_equal(
+                downside_risk,
+                expected,
+                DECIMAL_PLACES)
+        else:
+            for i in range(downside_risk.size):
+                assert_almost_equal(
```
llllllllll: you can give assert_almost_equal an array

setup.py
```diff
@@ -4,45 +4,60 @@
...
+test_reqs = [
+    "nose>=1.3.7",
+    "nose_parameterized>=0.5.0"
+]
+
 
-install_reqs = [
+requirements = [
     'numpy>=1.9.1',
     'pandas>=0.15.0',
```
llllllllll: this is acually pandas>=0.18 based on the features used.

qrisk/tests/test_stats.py
```diff
@@ -2,75 +2,154 @@
```
llllllllll: since this is a small repo and porting wouldn't be hard, I am wondering why we chose to use nose instead of pytest. nose is currently unsupported so I would prefer not to start a new project using it.

qrisk/stats.py
ahgnaw: You can't because if any of the returns is 0.0, then the result will be 0.0 overall. [.05, 0.0,] would yield 0. Adding 1 to the entire series allows the result to be 0.5, but forces the 1 to be subtracted at the end.

README.md
ahgnaw: It's only on the pypi test server right now, and then not the latest version

qrisk/stats.py
llllllllll: was the default just incorrectly documented then?

qrisk/stats.py
ahgnaw: We don't, that was an error that wasn't tested.

llllllllll: nvm, we have the division future imported, the `float(...)` threw me off though, it doesn't do anything here so we should probably remove it.

setup.py
ahgnaw: Travis is running with pandas 0.16.1, what is it missing?

qrisk/tests/test_stats.py
ahgnaw: Its what zipline and pyfolio are currently using.

qrisk/tests/test_stats.py
ahgnaw: Sweet. I assumed you couldn't.

qrisk/tests/test_stats.py
llllllllll: I don't think that is sufficient justification for continuing to use an unsupported library. The first sentence of the nose docs is:

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

setup.py
llllllllll: We are using the new resample api, have we run this on travis yet? The build hasn't run since there are merge conflicts.

qrisk/stats.py
ahgnaw: GP

qrisk/stats.py
ahgnaw: I don't have a preference either way, but why do you think len() is better than .size?

setup.py
llllllllll: actually, we might be fine with the way we are using resample. I would be very suprised if this actually worked with 0.15 though, it should be at least >= 0.16.1

qrisk/stats.py
ahgnaw: Ah yes, that is a poorly written sentence.

qrisk/stats.py
llllllllll: `len` is the standard way to talk about the length of some collection, `size` has an overloaded meaning in numpy/pandas, for example:

```
In [52]: len(np.arange(25).reshape(5, 5))
Out[52]: 5

In [53]: np.arange(25).reshape(5, 5).size
Out[53]: 25
```

I think it is just more standard

qrisk/stats.py
ahgnaw: Oh, I read that wrong. Yeah turning it into a function makes sense.

.travis.yml
ahgnaw: Nope, i'll remove it.

qrisk/__init__.py
ahgnaw: Yes, flake8 won't like all the unused imports though

qrisk/stats.py
ahgnaw: I'm not seeing where the docstring says the default is 1.

qrisk/stats.py
ahgnaw: I'm not very familiar with sphinx, would that be :func:`~qrisk.stats.cum_function` or :func:`~stats.cum_function`?

qrisk/stats.py
ahgnaw: Force of habit, I'll remove it

qrisk/__init__.py
llllllllll: then we can reference them ina `__all__`

qrisk/stats.py
llllllllll: In the diff, the docstring used to say: `(default 1)`

qrisk/stats.py
llllllllll: The first

qrisk/tests/test_stats.py
```diff
@@ -2,75 +2,154 @@
...
+            assert_almost_equal(
+                cum_returns[i],
+                expected[i],
+                4)
 
     @parameterized.expand([
-        (simple_rets, stats.DAILY, 0.15500998835658075),
-        (simple_week_rets, stats.WEEKLY, 0.030183329386562319),
-        (simple_month_rets, stats.MONTHLY, 0.006885932704891129)
+        (simple_benchmark, qrisk.WEEKLY, [0.010000000000000009,
```
llllllllll: are all these digits needed?

qrisk/stats.py
ahgnaw: That is an oversight, I'll change it.

qrisk/stats.py
ahgnaw: The default was incorrectly documented.

qrisk/stats.py
llllllllll:
```python
def max_drawdown(returns):
    if returns.size < 1:
        return np.nan

    end = (returns.cummax() - returns).argmax()
    return returns[end] - returns[:end].max()
```

this produces the same result but runs much faster

qrisk/stats.py
```diff
@@ -417,15 +432,13 @@ def sortino_ratio(returns, required_return=0, period=DAILY,
...
-    try:
-        ann_factor = annualization[period]
-    except KeyError:
-        raise ValueError(
-            "period cannot be '{}'. "
-            "Can be '{}'.".format(
-                period, "', '".join(annualization.keys())
-            )
-        )
+    if len(returns) < 2:
```
llllllllll: why do we check this twice?

PR #6, opened 2016-07-06
Changed index to multi index in aggregate returns function

ahgnaw:
Some tear sheet graphs in pyfolio require a multi index dataframe. Instead of returning a list of dates as an index, return a multi index that separates those dates by year, week number, month.

Aggregation function will use `Series.groupby` instead of `Series.resample`, which is currently deprecated.

### DIFF
--- a/qrisk/stats.py
+++ b/qrisk/stats.py
@@ -143,15 +143,17 @@ def aggregate_returns(returns, convert_to):
         return cum_returns(x)[-1]
 
     if convert_to == WEEKLY:
-        return returns.resample('W', how={'Agg': cumulate_returns})
+        grouping = [lambda x: x.year, lambda x: x.isocalendar()[1]]
     elif convert_to == MONTHLY:
-        return returns.resample('M', how={'Agg': cumulate_returns})
+        grouping = [lambda x: x.year, lambda x: x.month]
     elif convert_to == YEARLY:
-        return returns.resample('A', how={'Agg': cumulate_returns})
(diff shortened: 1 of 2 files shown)


qrisk/stats.py
```diff
@@ -143,15 +143,19 @@ def cumulate_returns(x):
         return cum_returns(x)[-1]
 
     if convert_to == WEEKLY:
-        return returns.resample('W', how={'Agg': cumulate_returns})
+        aggregate = returns.resample('W', how={'Agg': cumulate_returns})
+        aggregate.index = [aggregate.index.year, aggregate.index.week]
```
twiecki: Isn't there a `aggregate.set_index()` we can use for this?

qrisk/stats.py
```diff
@@ -143,16 +143,18 @@ def cumulate_returns(x):
...
-        return returns.resample('W', how={'Agg': cumulate_returns})
+        grouping = [lambda x: x.year, lambda x: x.isocalendar()[1]]
     elif convert_to == MONTHLY:
-        return returns.resample('M', how={'Agg': cumulate_returns})
+        grouping = [lambda x: x.year, lambda x: x.month]
     elif convert_to == YEARLY:
-        return returns.resample('A', how={'Agg': cumulate_returns})
+        grouping = [lambda x: x.year]
     else:
         ValueError(
```
abhijeetkalyan: Think this needs a `raise` before it?

abhijeetkalyan:
LGTM

PR #7, opened 2016-07-07
ENH: Annualize alpha calculation

ahgnaw:
The calculation of alpha should be annualized.d

### DIFF
--- a/qrisk/stats.py
+++ b/qrisk/stats.py
@@ -532,8 +532,9 @@ def information_ratio(returns, factor_returns):
     return np.mean(active_return) / tracking_error
 
 
-def alpha_beta(returns, factor_returns, risk_free=0.0):
-    """Calculates both alpha and beta.
+def alpha_beta(returns, factor_returns, risk_free=0.0, period=DAILY,
+               annualization=None):
+    """Calculates annualized alpha and beta.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


abhijeetkalyan:
Think it's better if @twiecki looks at this one

qrisk/stats.py
```diff
@@ -532,8 +532,9 @@ def information_ratio(returns, factor_returns):
     return np.mean(active_return) / tracking_error
 
 
-def alpha_beta(returns, factor_returns, risk_free=0.0):
-    """Calculates both alpha and beta.
+def alpha_beta(returns, factor_returns, risk_free=0.0, period=DAILY,
+               annualization=None):
```
twiecki: What's the default annualization factor?

twiecki:
The math looks correct to me.

qrisk/stats.py
ahgnaw: The default factor is 252 as it assumes daily returns.

jbredeche: Does it matter that even in a US equities only world, there aren't always
252 trading days in a year?

twiecki: The interaction between period being overwritten by annualization is a bit odd here and probably bears mentioning the doc-string.

@jbredeche Not a big deal here, it's common practice to just use 250 or 252.

ahgnaw: @jbredeche The default annualization can be be over written by `period`.

qrisk/stats.py
```diff
@@ -47,13 +47,14 @@ def annualization_factor(period, annualization):
...
+            'monthly':12
+            'weekly': 52
+            'daily': 252
     annualization : int, optional
-        Factor used to annualize returns when different from default values.
-        - Default:
-          annualization = {'daily': 252,
-                           'weekly': 52,
-                           'monthly': 12}
+        Used suppress default values available in `period` to convert returns
```
twiecki: missing "to".

PR #8, opened 2016-07-20
ENH: Use bottleneck to speed up calculations

ahgnaw:


twiecki:
Nice, didn't know about bottleneck before, looks really neat.

However, I'm a bit hesitant to merge this for the following reasons:
* Adds complexity to code base, installation, testing, dependency issues and ultimately support.
* It's not clear that at all that this is a bottleneck, the sizes over which these stats are computed are so small that numpy or pandas should be more than fast enough. What use case are we trying to optimize for?
* We're only using it for a few functions.

Let me know what you think.

jbredeche:
Out of curiosity, what kind of speedup do we get with bottleneck?

richafrank:
> Out of curiosity, what kind of speedup do we get with bottleneck?

Before: [`Ran 1430 tests in 908.394s`] (https://travis-ci.org/quantopian/zipline/jobs/145622689)
After: [`Ran 1440 tests in 581.959s`] (https://travis-ci.org/quantopian/zipline/jobs/145990277)

Though we should have looked at the performance earlier, the reason we made this change was that `numpy.nanmean` calls [`warnings.catch_warnings`] (https://docs.python.org/3.1/library/warnings.html#warnings.catch_warnings), which resets the warnings cache on python3.  This means that our deprecation warnings appear to the user every simulation day, whereas on python2 or with bottleneck, they only appear the first time the user is warned.

We were trying to prevent spamming the user with warnings.  We noticed this because our [unit tests] (https://travis-ci.org/quantopian/zipline/builds/145622688) actually check the number of warnings produced

setup.py
```diff
@@ -61,6 +61,7 @@
     'numpy>=1.9.2',
     'pandas>=0.16.1',
     'scipy>=0.15.1',
+    'bottleneck==1.0.0'
```
richafrank: Do we need to pin the exact version?

ahgnaw: Not at all

twiecki:
@richafrank so performance isn't the main objective here?

richafrank:
> @richafrank so performance isn't the main objective here?

It wasn't the original objective, but given the run times above, I think it's just as important.

twiecki:
OK, I'm in favor of this then, and the speed-up is meaningful.

PR #10, opened 2016-08-02
Alternative alphabeta

ahgnaw:
Property tests added to confirm functions behave as expected. Previous testing was limited to regression tests.

The calculation for alpha and beta was using linregress, it was unnecessary and caused a third of algorithms on performance matrix to slow down. The new way of calculating them is faster and correct.

### DIFF
--- a/qrisk/tests/test_stats.py
+++ b/qrisk/tests/test_stats.py
@@ -187,7 +187,8 @@ class TestStats(TestCase):
         else:
             assert constant*max_dd >= transformed_dd
 
-    # Translating returns by a positive constant should increase the maximum
+    # Maximum drawdown is always less than or equal to zero. Translating
+    # returns by a positive constant should increase the maximum
     # drawdown to a maximum of zero. Translating by a negative constant
     # decreases the maximum drawdown.
     @parameterized.expand([
@@ -303,12 +304,13 @@ class TestStats(TestCase):
             DECIMAL_PLACES)
 
     # Translating the returns and required returns by the same amount
-    # should not change the sharpe ratio.
+    # does not change the sharpe ratio.
     @parameterized.expand([
         (noise_uniform, 0, .005),
         (noise_uniform, 0.005, .005)
     ])
-    def test_sharpe_translation(self, returns, required_return, translation):
+    def test_sharpe_translation_same(self, returns, required_return,
+                                     translation):
         sr = qrisk.sharpe_ratio(returns, required_return)
         sr_depressed = qrisk.sharpe_ratio(
             returns-translation,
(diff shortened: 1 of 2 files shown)


brianpfink:
Now that you have stopped using linregress it is equally fast as before?

qrisk/tests/test_stats.py
```diff
@@ -127,6 +168,43 @@ def test_max_drawdown(self, returns, expected):
...
+        (noise_uniform, .1),
+    ])
+    def test_max_drawdown_translation(self, returns, constant):
+        depressed_returns = returns-constant
+        raised_returns = returns+constant
+        max_dd = qrisk.max_drawdown(returns)
+        depressed_dd = qrisk.max_drawdown(depressed_returns)
+        raised_dd = qrisk.max_drawdown(raised_returns)
+        assert max_dd <= raised_dd
+        assert depressed_dd <= max_dd
```
brianpfink: > Translating returns by a positive constant should increase the maximum drawdown to a maximum of zero
Can you add an assertion that the max drawdown is always less than 0?

> Translating by a negative constant decreases the maximum drawdown.
I think this only tests postive constants. Can you add tests for negative constants?

qrisk/tests/test_stats.py
```diff
@@ -283,21 +484,121 @@ def test_sortino(self, returns, required_return, period, expected):
...
+        returns[loss_loc[0]] = required_return
+        sr_2 = qrisk.sortino_ratio(returns, required_return)
+        returns[loss_loc[1]] = required_return
+        sr_3 = qrisk.sortino_ratio(returns, required_return)
+        assert sr_1 <= sr_2
+        assert sr_2 <= sr_3
+
+    # Translating the returns and required returns by the same amount
+    # should not change the sortino ratio.
+    @parameterized.expand([
```
brianpfink: Should we test that if we change it by different amounts sortino ratio does change?

brianpfink:
Nice job. This is a huge improvement to the excel sheet. It is great that we have test for alpha and beta so we do not need to worry if the performance change broke anything.

twiecki:
:+1:

ahgnaw:
@brianpfink Removing linregress speed up testing. Calling it once for beta and another time for alpha was a problem.

qrisk/tests/test_stats.py
ahgnaw: The test translates returns up and down by adding and subtracting to the constant. A test for negative constants would change the assertions at the end.

qrisk/tests/test_stats.py
```diff
@@ -210,9 +302,69 @@ def test_sharpe_ratio(self, returns, risk_free, expected):
             expected,
             DECIMAL_PLACES)
 
+    # Translating the returns and required returns by the same amount
+    # should not change the sharpe ratio.
+    @parameterized.expand([
+        (noise_uniform, 0, .005),
```
ahgnaw: gp, adding it

qrisk/tests/test_stats.py
ahgnaw: adding it

brianpfink:
Looks good to me!

PR #16, opened 2016-08-17
ENH: Remove dropna from beta function

ahgnaw:
Sometimes returns have an np.nan value which causes beta to return np.nan. In the case that either contains np.nan, calculate covariance of only those dates that have both returns and benchmark return values.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -659,13 +659,19 @@ def beta(returns, factor_returns, risk_free=0.0):
         Beta.
     """
 
-    if len(returns) < 2:
+    # Filter out dates with np.nan as a return value
+    indices = np.intersect1d(
+        returns.dropna().index,
+        factor_returns.dropna().index
+    )
+
+    if len(indices) < 2:
         return np.nan
 
-    covar = np.cov(returns.dropna()-risk_free,
-                   factor_returns.dropna(), ddof=0)[0][1]
+    covar = np.cov(returns[indices]-risk_free,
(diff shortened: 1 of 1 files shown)


twiecki:
```
======================================================================
FAIL: test_max_drawdown_transformation_1 (empyrical.tests.test_stats.TestStats)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/testenv/lib/python3.5/site-packages/nose_parameterized/parameterized.py", line 365, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/empyrical/empyrical/tests/test_stats.py", line 188, in test_max_drawdown_transformation
    assert constant*max_dd <= transformed_dd
AssertionError
======================================================================
FAIL: test_max_drawdown_transformation_4 (empyrical.tests.test_stats.TestStats)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/testenv/lib/python3.5/site-packages/nose_parameterized/parameterized.py", line 365, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/home/travis/build/quantopian/empyrical/empyrical/tests/test_stats.py", line 190, in test_max_drawdown_transformation
    assert constant*max_dd >= transformed_dd
AssertionError
```

These seem unrelated, but why are we getting them?

twiecki:
PR looks good to me. Would be good to have a test with nans here.

PR #18, opened 2016-08-19
TST: Include sparse input test for alpha and beta

ahgnaw:
This test ensures that when input contains np.nan values, alpha and beta can be calculated using the remaining data instead of returning np.nan.

### DIFF
--- a/empyrical/tests/test_stats.py
+++ b/empyrical/tests/test_stats.py
@@ -72,12 +72,12 @@ class TestStats(TestCase):
     flat_line_0 = pd.Series(
         np.linspace(0, 0, num=1000),
         index=pd.date_range('2000-1-30', periods=1000, freq='D')
-        )
+    )
     # Flat line
     flat_line_1 = pd.Series(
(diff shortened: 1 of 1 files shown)


empyrical/tests/test_stats.py
```diff
@@ -764,6 +772,16 @@ def test_alpha_beta_correlation(self, corr_less, corr_more):
...
         assert beta_less < beta_more
 
+    # When faced with data containing np.nan, do not return np.nan. Calculate
+    # alpha and beta using dates containing both.
+    @parameterized.expand([
+        (sparse_noise, sparse_flat_line_1),
+    ])
+    def test_alpha_beta_with_nan_inputs(self, returns, benchmark):
+        alpha, beta = empyrical.alpha_beta(returns, benchmark)
+        assert alpha != np.nan
```
twiecki: `self.assertNotEqual()` (or something like that)

ahgnaw: I've made the change.

twiecki:
:+1:

Issue #9, opened 2016-07-26
Add SQN metric


twiecki:
http://backtrader.readthedocs.io/en/latest/analyzers-reference.html#sqn

ahgnaw:
Interesting metric, I've never seen it before. Input can be a list or dataframe of closed trades, what do you think?

twiecki:
Yeah, specifying trades is actually not a straight-forward concept. In pyfolio we have `round_trip.py` for that which we could use to create it. But yeah, that list/dataframe as input here sounds good.

PR #19, opened 2016-08-19
ENH: Add CAGR metric with testing

ahgnaw:

PR #21, opened 2016-08-23
ENH: Add SQN calculation

LotannaEzenwa:
closes https://github.com/quantopian/empyrical/issues/9

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -770,6 +770,36 @@ def cagr(returns, period=DAILY, annualization=None):
     return ending_value ** (1. / no_years) - 1
 
 
+def SQN(trade_returns):
+    """Determines the SystemQualityNumber of trades as defined by Van K. Tharp.
+
+    Parameters
+    ----------
+    trade_returns : pd.Series
+        A series of transaction gains/losses.
+
+    Returns
+    -------
+    float
+        SystemQualityNumber
+
+    """
+
+    n_trades = len(trade_returns)
+
+    if n_trades < 30:
+        return np.nan
+
+    pos = trade_returns[trade_returns > 0]
+    std = pos.std()
+    avg = pos.mean()
+
+    if std == 0:
+        return np.nan
+
+    return np.sqrt(n_trades)*avg/std
+
+
 SIMPLE_STAT_FUNCS = [
     annual_return,
     annual_volatility,
@@ -789,4 +819,8 @@ FACTOR_STAT_FUNCS = [
     information_ratio,
     alpha,
     beta,
+]
+
+TRADE_STAT_FUNCS = [
+    SQN,
 ]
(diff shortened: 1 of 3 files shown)


empyrical/stats.py
```diff
@@ -727,6 +727,39 @@ def tail_ratio(returns):
...
+
+    pos = trade_returns[trade_returns > 0]
+
+    if len(pos) < 3:
+        return np.nan
+
+    std = np.nanstd(pos, ddof=1)
+    avg = np.nanmean(pos)
+
+    if abs(std) < np.finfo(np.float).eps:
```
llllllllll: isn't this just `std == 0`?

empyrical/stats.py
```diff
@@ -727,6 +727,39 @@ def tail_ratio(returns):
...
+        return np.nan
+
+    return np.sqrt(n_trades)*avg/std
```
llllllllll: please put spaces around binary operators

empyrical/stats.py
```diff
@@ -727,6 +727,39 @@ def tail_ratio(returns):
         np.abs(np.percentile(returns, 5))
 
 
+def SQN(trade_returns):
```
llllllllll: is there a reason that this is capitalized?

empyrical/tests/test_stats.py
```diff
@@ -837,3 +839,17 @@ def test_tail_ratio(self, returns, expected):
             empyrical.tail_ratio(returns),
             expected,
             1)
+
+    @parameterized.expand([
+        (empty_returns, np.nan),
+        (one_return, np.nan),
+        (flat_line_1, np.nan),
+        (pos_line, 54.799662435119096),
```
llllllllll: can you explain how these these inputes/outputs were chosen?

empyrical/stats.py
LotannaEzenwa: At first I ran into some errors with precision, so that was added. Added it for a sanity check. Changing it back to std==0..

empyrical/stats.py
LotannaEzenwa: 👍

empyrical/stats.py
LotannaEzenwa: other than being an abbreviation, no.
it looks like CAGR is also being defined as `cagr` in #19

empyrical/tests/test_stats.py
LotannaEzenwa: 0-1. return `nan` when not enough transactions passed in
2. return `nan` when the std is zero
3. the positive line captures the correct functionality. It removes the first value, 0, and computes the mean and std of the rest of the line, then returns the value.
4. negative line should return `nan` as none of the values are positive.

Should there be additional/different tests? I'd imagine it could be improved with real data

LotannaEzenwa:
@ahgnaw there's that repro 😆

gusgordon:
Closing this as it's old. Feel free to reopen if this is still relevant.

PR #25, opened 2016-08-25
BUG: Beta works with sparse timezoned input

ahgnaw:
Instead of returning nan when input is sparse, beta now returns a value using available data.

PR #26, opened 2016-08-26
WIP Bug 23 fix2

twiecki:
Continuation of https://github.com/quantopian/empyrical/pull/25

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -706,15 +706,12 @@ def beta(returns, factor_returns, risk_free=0.0):
     if len(returns) < 2 or len(factor_returns) < 2:
         return np.nan
     # Filter out dates with np.nan as a return value
-    indices = pd.concat([returns, factor_returns], axis=1).dropna()
-    if len(indices) < 2:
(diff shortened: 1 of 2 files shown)


twiecki:
I think a constant array is a bad test-case for the factor_returns, I changed them to something else, can still think about that. But it does require the numbers in the tests to be adapted.

jdricklefs:
@jbredeche 

Using the 1000-item random-noise series from the unit tests

```
In [6]: %timeit new_beta(noise, noise, 0.0)
1000 loops, best of 3: 1.09 ms per loop

In [8]: %timeit old_beta(noise, noise, 0.0)
1000 loops, best of 3: 1.02 ms per loop
```

Doesn't seem significantly different.

jdricklefs:
> I think a constant array is a bad test-case for the factor_returns,

In Quantopian-terms, it seems possible that an algorithm could set their benchmark to a relatively illiquid security, resulting in some duration of constant returns? Is this something we should handle or is it an inherently broken scenario?

jbredeche:
thanks @jdricklefs

PR #28, opened 2016-08-30
WIP Beta should use slope, not correlation. Improve unittests to also…

twiecki:
… compare against scipy linregress. Some tests still need to be fixed.


twiecki:
The main problem was that I used `.corr()` while beta is supposed to refer to the slope between https://github.com/quantopian/empyrical/commit/28b59b6ee7f8356b1bda24aadd6e03fa6557a9b6 This uses the indexing fix but reverts back to the original cov/var calculation and improves the tests a bit.

PR #29, opened 2016-08-30
Add tz testcases

ahgnaw:
Test functions using timezone aware data series.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -711,6 +711,9 @@ def beta(returns, factor_returns, risk_free=0.0):
     if len(joint) < 2:
         return np.nan
 
+    if np.absolute(joint.var().iloc[1]) < 1.0e-30:
+        return np.nan
+
     return joint.cov().iloc[0, 1] / joint.var().iloc[1]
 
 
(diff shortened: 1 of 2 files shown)


empyrical/stats.py
```diff
@@ -714,7 +714,7 @@ def beta(returns, factor_returns, risk_free=0.0):
     if np.absolute(joint.var().iloc[1]) < 1.0e-30:
         return np.nan
 
-    return joint.cov().iloc[0, 1] / joint.var().iloc[1]
+    return np.cov(joint.values.T, ddof=0)[0, 1] / np.var(joint.iloc[:, 1])
```
ahgnaw: Speed increase from switching to np.cov instead of cov() confirmed:
    In [32]: %paste
    start = time.time()
    for i in range(num_trials):
            tmp = joint.cov().iloc[0, 1]
    pointA = time.time()
    for i in range(num_trials):
            tmp = np.cov(joint.values.T, ddof=0)[0, 1]
    pointB = time.time()
    
    print(pointA - start)
    print(pointB - pointA)
    
    ## -- End pasted text --
    0.6068220138549805
    0.06141209602355957

PR #30, opened 2016-08-31
MAINT Beta calculation does not require explicity variance calculation.

twiecki:
Variance is already computed for the covariance matrix, can just use that.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -711,10 +711,11 @@ def beta(returns, factor_returns, risk_free=0.0):
     if len(joint) < 2:
         return np.nan
 
-    if np.absolute(joint.var().iloc[1]) < 1.0e-30:
+    cov = np.cov(joint.values.T, ddof=0)
+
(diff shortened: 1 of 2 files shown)


empyrical/stats.py
```diff
@@ -714,7 +714,8 @@ def beta(returns, factor_returns, risk_free=0.0):
     if np.absolute(joint.var().iloc[1]) < 1.0e-30:
         return np.nan
 
-    return np.cov(joint.values.T, ddof=0)[0, 1] / np.var(joint.iloc[:, 1])
+    cov = np.cov(joint.values.T, ddof=0)
```
ssanderson: The line above this is doing `np.absolute(joint.var().iloc[1])`.  Are we duplicating work here?

ahgnaw: Yes, and there's a pr to reuse the covariance matrix values
https://github.com/quantopian/empyrical/pull/30/commits/19cc30e16b32bae4797ff56079e967a2e14ba172

PR #31, opened 2016-09-13
TST: Change noop assertions to check for nan

richafrank:
(Everything is not equal to nan!)

### DIFF
--- a/empyrical/tests/test_stats.py
+++ b/empyrical/tests/test_stats.py
@@ -800,7 +800,7 @@ class TestStats(TestCase):
     # When faced with data containing np.nan, do not return np.nan. Calculate
     # alpha and beta using dates containing both.
     @parameterized.expand([
-        (sparse_noise, sparse_flat_line_1_tz),
+        (sparse_noise, sparse_noise),
     ])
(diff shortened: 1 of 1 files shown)


richafrank:
@twiecki I changed these assertions from being trivially true, since everything evaluates not equal to nan, and now one of them fails.  I'm not sure if the test or the risk function(s) should be changed.  What are your thoughts?

richafrank:
Specifically, the test that fails is `test_alpha_beta_with_nan_inputs_0`.

twiecki:
I think the problem is not the nans in the input here but instead the flat-line we pass in for benchmark. @ahgnaw can we replace all of those with non-flat lines? I'm not sure what relevant case we're testing there.

ahgnaw:
The flat line here is  wrongly used as an input. The alpha and beta values of any return series against a flat line will return np.nan. In other tests, the flat line is used as a sanity check, e.g. `test_alpha_2`. Getting rid of all tests that use a flat line as a benchmark would remove some property tests.

ahgnaw:
LGTM

twiecki:
:+1:

PR #32, opened 2016-09-16
Enumpycal

richafrank:
Main change:

PERF: Enable stats to work with Series or ndarrays
Speeds up zipline, which can use ndarrays

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -43,8 +43,8 @@ ANNUALIZATION_FACTORS = {
 
 def _adjust_returns(returns, adjustment_factor):
     """
-    Returns a new pd.Series adjusted by adjustment_factor. Optimizes for the
-    case of adjustment_factor being 0
+    Returns the returns series adjusted by adjustment_factor. Optimizes for the
+    case of adjustment_factor being 0 by returning returns itself, not a copy!
 
     Parameters
     ----------
@@ -56,7 +56,7 @@ def _adjust_returns(returns, adjustment_factor):
     pd.Series
     """
     if isinstance(adjustment_factor, (float, int)) and adjustment_factor == 0:
-        return returns.copy()
+        return returns
     return returns - adjustment_factor
 
 
@@ -540,7 +540,7 @@ def downside_risk(returns, required_return=0, period=DAILY,
 
     ann_factor = annualization_factor(period, annualization)
 
-    downside_diff = _adjust_returns(returns, required_return)
+    downside_diff = _adjust_returns(returns, required_return).copy()
     mask = downside_diff > 0
     downside_diff[mask] = 0.0
     squares = np.square(downside_diff)
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -462,7 +462,7 @@ def sharpe_ratio(returns, risk_free=0, period=DAILY, annualization=None):
 
     """
 
-    if len(returns) < 1:
+    if len(returns) < 2:
         return np.nan
 
     ann_factor = annualization_factor(period, annualization)
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -138,7 +138,7 @@ def cum_returns(returns, starting_value=0):
     # to use.
 
     if len(returns) < 1:
-        return [np.nan]
+        return type(returns)([])
 
     if np.isnan(returns[0]):
         returns = returns.copy()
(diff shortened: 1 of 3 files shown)


empyrical/stats.py
```diff
@@ -43,20 +46,20 @@
...
+    adjustment_factor : pd.Series or np.ndarray or float or int
 
     Returns
     -------
-    pd.Series
+    pd.Series or np.ndarray
     """
     if isinstance(adjustment_factor, (float, int)) and adjustment_factor == 0:
-        return returns.copy()
+        return returns
```
jdricklefs: Hey, nice. 👍

empyrical/tests/test_stats.py
```diff
@@ -955,3 +971,159 @@ def test_cagr_noisy(self, returns, add_noise):
...
+        newone.__dict__.update(self.__dict__)
+        return newone
+
+    def __getattr__(self, item):
+        return self._check_return_type(getattr(empyrical, item))
+
+    def _check_return_type(self, func):
+        @wraps(func)
+        def check_return_type(*args, **kwargs):
+            result = func(*args, **kwargs)
```
jdricklefs: Would be pretty cool to extend this to validate that the input data was not mutated (given that we also removed my paranoid `returns.copy()` in `_adjust_returns`).

empyrical/tests/test_stats.py
```diff
@@ -955,3 +971,159 @@ def test_cagr_noisy(self, returns, add_noise):
...
+    def empyrical(self):
+        return PassArraysEmpyricalProxy(self, (np.ndarray, float))
+
+
+class TestStatsIntIndex(TestStats):
+    """
+    Tests pass int-indexed pd.Series inputs to empyrical and assert that
+    outputs are of type pd.Series or float.
+
+    This prevents a regression where we're indexing with ints into a pd.Series
```
jdricklefs: 👍

empyrical/tests/test_stats.py
```diff
@@ -955,3 +971,159 @@ def test_cagr_noisy(self, returns, add_noise):
             cagr,
             noisy_cagr_2,
             1)
+
+    @property
+    def empyrical(self):
+        """
+        This returns a wrapper around the empyrical module, used a hook, so
```
jdricklefs: The sentence structure here is pretty rough. 

Maybe something like ...

> This returns a wrapper around the empyrical module which allows the test-author to specify input conversions or return type-checks around each call to an empyrical function.

empyrical/tests/test_stats.py
```diff
@@ -955,3 +971,159 @@ def test_cagr_noisy(self, returns, add_noise):
...
+        empyrical
+
+        Notes
+        -----
+        Since some parameterized test parameters refer to attributes on the
+        real empyrical module at class body scope, this property must be
+        defined later in the body than those references. That way, the
+        attributes are looked up on the empyrical module, not this property.
+
+        e.g. empyrical.DAILY
```
jdricklefs: Would it make sense to alias the import so this is a little more explicit?

richafrank: Could certainly do that.  Was trying to reduce the diff to show that the tests didn't change materially, but maybe it's confusing this way?

empyrical/tests/test_stats.py
richafrank: Thanks, will fix.

empyrical/tests/test_stats.py
richafrank: Done.

empyrical/stats.py
```diff
@@ -702,16 +906,49 @@ def beta(returns, factor_returns, risk_free=0.0):
...
+        Beta.
+    """
 
     if len(returns) < 2 or len(factor_returns) < 2:
         return np.nan
     # Filter out dates with np.nan as a return value
-    joint = pd.concat([_adjust_returns(returns, risk_free),
-                       factor_returns], axis=1).dropna()
-    if len(joint) < 2:
+    joint = np.vstack([_adjust_returns(returns, risk_free),
```
ahgnaw: `np.vstack` requires input arrays to be the same size, which is not always the case.

richafrank: Ah, but this is in `beta_aligned` now, so we expect them to be the same size already.  zipline's arrays are already aligned, but if a consumer can't guarantee that, they should use the existing [`beta` function] (https://github.com/quantopian/empyrical/pull/32/files/4d24d3f551e76b6027ceb804b4d1d5362d1e5b68#diff-23342a2bf4f4c31bfe2dfe0bcbe54c8bR912), which calls this after aligning them.  `beta` is what pyfolio still calls.

Does that sound ok to you?

empyrical/stats.py
```diff
@@ -739,11 +976,11 @@ def stability_of_timeseries(returns):
     if len(returns) < 2:
         return np.nan
 
-    returns = returns.dropna()
+    returns = returns[~np.isnan(returns)]
```
ahgnaw: On my computer, `dropna` has better performance than this other method, is there a reason for changing this line?

empyrical/stats.py
ahgnaw: Ooo, gp gp.

empyrical/stats.py
richafrank: Oh interesting.  I changed it to use a method that supports both `Series` and `ndarray,` but looks like it's slower on `Series` than `dropna` because `dropna` first converts to using an array.  I'll do the same here.

ahgnaw:
LGTM

empyrical/stats.py
richafrank: Done.

twiecki:
Looks great! :shipit:

empyrical/tests/test_stats.py
```diff
@@ -955,3 +971,200 @@ def test_cagr_noisy(self, returns, add_noise):
...
+            arg_copies = [(i, arg.copy()) for i, arg in enumerate(args)
+                          if isinstance(arg, (NDFrame, np.ndarray))]
+            kwarg_copies = {
+                k: v.copy() for k, v in iteritems(kwargs)
+                if isinstance(v, (NDFrame, np.ndarray))
+            }
+
+            result = func(*args, **kwargs)
+
+            # Check that inputs weren't mutated by func.
```
jdricklefs: Woot, thanks!

PR #33, opened 2016-09-30
ENH: Add total_return function

ahgnaw:


empyrical/stats.py
```diff
@@ -152,6 +152,34 @@ def cum_returns(returns, starting_value=0):
...
+
+    Returns
+    -------
+    float
+
+
+
+    """
+
+    return cum_returns(returns, starting_value).iloc[-1]
```
richafrank: If we pass an ndarray, then there's no `iloc` attribute, right?  

Want to add a test to the suite?  If you add it to the base class, the subclasses will try with different input types.

empyrical/stats.py
```diff
@@ -152,6 +152,40 @@ def cum_returns(returns, starting_value=0):
...
+
+    if len(returns) == 0:
+        return np.nan
+
+    cumulative_returns = cum_returns(returns, starting_value)
+
+    if isinstance(cumulative_returns, np.ndarray):
+        return cumulative_returns[-1]
+    else:
+        return cum_returns(returns, starting_value).iloc[-1]
```
richafrank: Looks like in this case we calculate the `cum_returns` twice!

I think, instead of doing an instance check in order to know the return type of `cum_returns` so we can get the last item, it's likely much faster to always pull the array out of the series and pass that through.  Then you'll also know how to index into the returned array.  (See https://github.com/quantopian/empyrical/blob/1dc28d51b16a55350c09ca60d1b3c2ed4be78ee5/empyrical/stats.py#L334-L336)

abhijeetkalyan:
I don't think there's anything for me to do here 😄

twiecki:
LGTM, the name is not perfect but I can't really think of a better one. Here are a few ideas: `cum_returns_final`, `cum_returns_total`, `cum_returns_terminal`.

ahgnaw:
@twiecki I like `cum_returns_final` and will change the function name.

PR #37, opened 2016-11-28
ENH: Use nanstd and nanmean, remove skipped test

ahgnaw:
Use nanmean and nanstd instead of np.mean and np.std to handle data sets with missing values.

A skipped test with poor reasoning is removed.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -558,10 +558,10 @@ def sharpe_ratio(returns, risk_free=0, period=DAILY, annualization=None):
     returns_risk_adj = np.asanyarray(_adjust_returns(returns, risk_free))
     returns_risk_adj = returns_risk_adj[~np.isnan(returns_risk_adj)]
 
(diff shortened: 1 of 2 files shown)


empyrical/stats.py
```diff
@@ -558,10 +558,10 @@ def sharpe_ratio(returns, risk_free=0, period=DAILY, annualization=None):
     returns_risk_adj = np.asanyarray(_adjust_returns(returns, risk_free))
     returns_risk_adj = returns_risk_adj[~np.isnan(returns_risk_adj)]
 
-    if np.std(returns_risk_adj, ddof=1) == 0:
+    if nanstd(returns_risk_adj, ddof=1) == 0:
```
richafrank: When would nans appear here?  I think the line above this filters them out.

richafrank: Ah cool - I thought this was a change in behavior.

twiecki:
Going to merge these.

PR #38, opened 2016-12-02
ENH Allow cum_returns to accept DataFrame

gusgordon:
In order for `cum_returns` to correctly support vector input, we need to check all columns using `np.all`.

Before, we were getting this error when passing some vector returns:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
cc @twiecki

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -140,7 +140,7 @@ def cum_returns(returns, starting_value=0):
     if len(returns) < 1:
         return type(returns)([])
 
-    if np.all(np.isnan(np.asanyarray(returns)[0])):
+    if np.isnan(np.asanyarray(returns)[0]):
         returns = returns.copy()
         returns[0] = 0.
 


ssanderson:
@gusgordon what do you mean by "vector input" here?  Are we expecting this function to vectorize over 2-d inputs?  If so, we should probably be testing that explicitly somewhere.

twiecki:
@ssanderson Yes, we need that for the cones that pass in an array of returns. Agreed on tests.

ssanderson:
> Yes, we need that for the cones that pass in an array of returns.

I'm confused because this already supports passing an array of returns.  Do you mean that we have cones that are passing multiple separate return streams and expecting this to be applied across all the different streams?

gusgordon:
@ssanderson Right. We have multiple columns of returns streams that we want to cumulate.

ssanderson:
Gotcha.  That makes sense.  We should also update the docstring here to reflect the updated types of expected inputs/outputs.

gusgordon:
I'm forgoing tests for the 2d case because empyrical's entire test structure has a wrapper that makes sure tested functions return arrays or floats. I don't see an easy way around this.

ssanderson:
cc @richafrank re: the issue @gusgordon is running into

richafrank:
Talked with @gusgordon IRL about making a new pair of `TestCase`s for 2D inputs.

gusgordon:
@richafrank Can you review? Thanks for the help!

richafrank:
Can do!

empyrical/stats.py
```diff
@@ -140,11 +142,14 @@ def cum_returns(returns, starting_value=0):
...
-        returns[0] = 0.
+    if isinstance(returns, pd.DataFrame):
+        df_cum = (returns.fillna(0) + 1).cumprod()
 
-    df_cum = np.exp(nancumsum(np.log1p(returns)))
+    else:
+        if np.all(np.isnan(np.asanyarray(returns)[0])):
+            returns = returns.copy()
+            returns[0] = 0.
+        df_cum = np.exp(nancumsum(np.log1p(returns)))
```
twiecki: Can't this run on the DataFrame?

richafrank: Agreed, I was going to suggest this too.  If we can use the same code for 1D and 2D+ arrays, then we don't need to test them both as essentially isolated functions.  I think we need to pass `axis=0` to `nancumsum` here.  We also need to fix the assignment to the zero column on the line above, in the case of a DataFrame.  

Would also be nice if this isn't DataFrame-specific, as opposed to an ndarray with more than 1 dimension.

empyrical/stats.py
```diff
@@ -140,11 +142,14 @@ def cum_returns(returns, starting_value=0):
...
 
-    if np.isnan(np.asanyarray(returns)[0]):
-        returns = returns.copy()
-        returns[0] = 0.
+    if isinstance(returns, pd.DataFrame):
+        df_cum = (returns.fillna(0) + 1).cumprod()
 
-    df_cum = np.exp(nancumsum(np.log1p(returns)))
+    else:
+        if np.all(np.isnan(np.asanyarray(returns)[0])):
```
richafrank: If one of the returns streams has a nan at the start, do we not want to replace it with a zero, since this is the case for a 1D array?

richafrank: Are you saying even the existing nan check doesn't belong here?

twiecki: Sorry, no. This is for the case where a user passes in either a `pd.Series().pct_change()` or a `pd.DataFrame().pct_change()` of prices. The current `all()` catches both of those cases. `any()`, as I understood you were proposing, would catch a case that we do not anticipate.

PR #39, opened 2016-12-05
ENH Allow cum_returns to accept DataFrame

gusgordon:
Adds support for passing a `DataFrame` to `stats.cum_returns`. Adds related tests.

I messed up the commits, so I made a new PR; sorry about that.

cc @richafrank @twiecki


empyrical/tests/test_stats.py
```diff
@@ -1019,6 +1019,62 @@ def empyrical(self):
...
+        for column in returns.columns:
+            for i in range(len(returns[column])):
+                assert_almost_equal(
+                    cum_returns[column][i],
+                    expected[column][i],
+                    4)
+
+    @property
+    def empyrical(self):
+        return ReturnTypeEmpyricalProxy(self, (pd.DataFrame))
```
richafrank: Should we add subclasses that test the other inputs we (I assume) now support, like `ndarray`s of more than 1 dimension?

richafrank: Not a big deal - the extra parens here aren't necessary.

empyrical/tests/test_stats.py
```diff
@@ -1019,6 +1019,62 @@ def empyrical(self):
...
+
+        for column in returns.columns:
+            for i in range(len(returns[column])):
+                assert_almost_equal(
+                    cum_returns[column][i],
+                    expected[column][i],
+                    4)
+
+    @property
+    def empyrical(self):
```
richafrank: Could you do me a favor and add a docstring here that says it returns "empyrical", so that my dev env will autocomplete empyrical's functions for `self.empyrical`?  Would be good to mention what we're testing with this property.

empyrical/stats.py
```diff
@@ -140,11 +142,11 @@ def cum_returns(returns, starting_value=0):
     if len(returns) < 1:
         return type(returns)([])
 
-    if np.isnan(np.asanyarray(returns)[0]):
+    if np.any(np.isnan(np.asanyarray(returns)[0])):
```
twiecki: We should only catch the case where the whole first row are `nan`s, as it's the only case produced by `.pct_change()`.

empyrical/tests/test_stats.py
```diff
@@ -1048,6 +1048,90 @@ def empyrical(self):
...
+                assert_almost_equal(
+                    cum_returns[column][i],
+                    expected[column][i],
+                    4)
+
+    @parameterized.expand([
+        (df_input.as_matrix(), 0, df_0_expected.as_matrix()),
+        (df_input.as_matrix(), 100, df_100_expected.as_matrix())
+    ])
+    def test_cum_returns_matrix(self, returns, starting_value, expected):
```
richafrank: Instead of duplicating the test methods for the input type, can we reuse the `EmpyricalProxy` machinery to add a subclass?
```python
@property
def empyrical(self):
    return PassArraysEmpyricalProxy(self, np.ndarray)
```

If we have one class that just deals with DataFrames and another with ndarrays, then each can check that the output type matches the input type.  Right now, we're asserting that DataFrame input will come back as either a DataFrame or an ndarray (and same with ndarray).

empyrical/tests/test_stats.py
```diff
@@ -1048,6 +1048,90 @@ def empyrical(self):
         )
 
 
+class Test2DStats(TestCase):
+    """
+    Tests for functions that are capable of outputting a DataFrame.
+    """
+
+    input_one = [np.nan, 0.01322056, 0.03063862, -0.01422057,
+                 -0.00489779, 0.01268925, -0.03357711, 0.01797036]
```
richafrank: Should we add the edge cases that we have for 1D, like empty and nans in the final position?

empyrical/stats.py
```diff
@@ -140,11 +142,11 @@ def cum_returns(returns, starting_value=0):
     if len(returns) < 1:
         return type(returns)([])
 
-    if np.isnan(np.asanyarray(returns)[0]):
+    if np.any(np.isnan(returns)):
```
twiecki: I don't feel deeply about this, it's really a corner case.

empyrical/tests/test_stats.py
```diff
@@ -1048,6 +1048,83 @@ def empyrical(self):
...
+
+        """
+
+        return ReturnTypeEmpyricalProxy(self, pd.DataFrame)
+
+
+class Test2DStatsArrays(Test2DStats):
+    """
+    Tests pass np.ndarray inputs to empyrical and assert that outputs are of
+    type np.ndarray or float.
```
richafrank: Looks like copy/paste mistake - No float!

empyrical/stats.py
richafrank: Ok, I wasn't sure, if we start seeing nans at arbitrary locations, whether we should mask it or raise an error or something else.

richafrank: My preference is to alert early instead of masking the unexpected input.

gusgordon:
Thanks @richafrank! We were doing that before anyway, and I think if we want to just replace the NaNs in the first row we would have to break apart the logic for different types.

gusgordon:
I opened this issue https://github.com/quantopian/empyrical/issues/40

Issue #46, opened 2017-05-18
Add Additional Calculations & Rolling Periods


cgdeboer:
@twiecki and friends, I'm working on the project that requires many of the same calculations that are performed in this repo, but also several others.  I am starting to work on a PR to add a variety of other statistical measures as well as a framework for getting rolling metrics on those things. 

Here are a few examples of additional metrics: `up-capture`, `down-capture`, `tracking-error`.

Would like to add some ability to create rolling metrics. eg: `Rolling 60-day Max Drawdown`, `Rolling 12 month Tracking Error` 

Before I start this work in earnest, I want to ensure that you all are open to expanding the calculation and capability set. Also, I can include a correct calculation of information ratio.

twiecki:
@cgdeboer That would be much appreciated and we'd be more than happy to work with you on this.

cgdeboer:
Great ! I'll post an initial PR tomorrow or early next week as a "Work In Progress" so I can get some initial thoughts from  you.

PR #47, opened 2017-05-19
Additional Calculations & Roll, Up, Down Helpers

### LINKED ISSUE #41: Remove Information Ratio
Our Information Ratio calculation (https://github.com/quantopian/empyrical/blob/master/empyrical/stats.py#L590) is wrong as noted by @marketneutral:
"The definition is wrong. You need a risk model to calc active risk to benchmark. And the benchmark look through to calc active return."

As this is still a ways off, we should rather delete it instead of having a wrong measure in here.

cgdeboer:
## Summary
Added a few helper methods in the utils file that allow us to stack some transformation logic together before calling an actual metric. For example, if we want to calculate the rolling up capture of a strategy vs its benchmark, we want to be able to do something like `[roll,up, capture]` where `roll`, `up`, `capture`, are each methods that perform either a series deconstruction (roll), a filter (up), or a calculation (capture). Doing it this why is more DRY than embedding roll and filter logic in each method, but does add some complication. 

I elected to add a variety of small methods to the `stats` file that wrap the complexity described above. This way we keep the API the user is used to calling the same. e.g there is actually a method called `rolling_up_capture` that takes the expected "returns, factor_returns" arguments. That function simply calls the more complex stacked function to get the results.

## Details
- added new methods details below.
- added roll, up, and down as stackable methods to use in future calculations
- added a handful of tests for new methods
- corrected an issue with the `max_drawdown`. See note below
- updated `nose_parameterized` to just `parameterized`
- move date/period symbols into a separate file
- added some examples to the README

## Not Done
- add `information_ratio` per #41 
- bump `empyrical` version
- add remainder of tests.

#### Max Drawdown
The `max_drawdown` code was not considering the first return in the array. The `max_return` array did not consider the starting value of 100 used in the function. In cases where ALL the returns in the array are negative, the max value in the array will be the 100 * (1+ the first negative return). For e.g if the first return in the array is `-0.02`, then the max value in the array will be 98. This will cause max drawdowns to appear less severe than if they started at 100. 

The max drawdown code was masquerading as correct because the `negative_returns` Series in the test suite starts with a 0. (i.e they aren't all negative). 

I've corrected that by ensuring that the `cum_returns` array starts with a value that matches the base value.

#### New Methods
```
    capture,
    up_capture,
    down_capture,
    up_down_capture,
    up_alpha_beta,
    down_alpha_beta,
    roll_max_drawdown,
    roll_up_capture,
    roll_down_capture,
    roll_up_down_capture,
    roll_alpha_beta,
    roll_sharpe_ratio,
```

Resolves #41 #46


empyrical/stats.py
```diff
@@ -587,7 +572,7 @@ def downside_risk(returns, required_return=0, period=DAILY,
     return dside_risk
 
 
-def information_ratio(returns, factor_returns):
+def excess_sharpe(returns, factor_returns):
     """
     Determines the Information ratio of a strategy.
```
twiecki: need to change doc string.

empyrical/stats.py
```diff
@@ -1004,6 +990,208 @@ def cagr(returns, period=DAILY, annualization=None):
 
     return ending_value ** (1. / no_years) - 1
 
+def capture(returns, factor_returns, period=DAILY):
+    """
```
twiecki: needs a reference

twiecki:
@cgdeboer I won't do a full review yet as you label it as WIP but wanted to say that this looks fantastic.

cgdeboer:
@twiecki I'll get this thing linted etc, and address your initial comments above. I'll remove the WIP title and notify you here once that's done.

empyrical/stats.py
cgdeboer: What kind of reference ? 

- http://www.morningstar.com/InvGlossary/upside-downside-capture-ratio.aspx

- http://www.investopedia.com/terms/u/up-market-capture-ratio.asp

Something like that? or am I misunderstanding what you mean by reference ?

cgdeboer:
@twiecki alrighty, I still have some tests to write and verify, and I'm still conferring with some colleagues on the info ratio stuff (i.e how to best proceed with that). At my own peril, I'm removing the WIP, review away.

cgdeboer:
@twiecki et al, let me know what you think. Would love to get this merged in this week or next.

twiecki:
@cgdeboer will take a look soon!

setup.py
```diff
@@ -61,7 +61,6 @@
     'numpy>=1.9.2',
     'pandas>=0.16.1',
     'scipy>=0.15.1',
-    'bottleneck>=1.0.0'
```
twiecki: Why?

empyrical/stats.py
```diff
@@ -720,11 +708,11 @@ def alpha_beta_aligned(returns, factor_returns, risk_free=0.0, period=DAILY,
...
     Returns
     -------
-    float
-        Alpha.
-    float
-        Beta.
-
+    list
+        float
+            Alpha.
```
twiecki: mixed tabs and spaces?

empyrical/stats.py
```diff
@@ -1005,6 +993,214 @@ def cagr(returns, period=DAILY, annualization=None):
...
+        annualizing. Value ignored if `annualization` parameter is specified.
+        Defaults are:
+            'monthly':12
+            'weekly': 52
+            'daily': 252
+    Returns
+    -------
+    float, np.nan
+        The capture ratio.
+    """
```
twiecki: Can you add links to http://www.investopedia.com/terms/u/up-market-capture-ratio.asp ?

empyrical/stats.py
```diff
@@ -1018,11 +1214,12 @@ def cagr(returns, period=DAILY, annualization=None):
     stats.skew,
     stats.kurtosis,
     tail_ratio,
-    cagr
+    cagr,
+    capture
```
twiecki: why not add up and down captures here?

empyrical/utils.py
```diff
@@ -24,12 +26,63 @@
...
     nanmean = np.nanmean
     nanstd = np.nanstd
     nansum = np.nansum
     nanmax = np.nanmax
     nanmin = np.nanmin
     nanargmax = np.nanargmax
     nanargmin = np.nanargmin
+
+
+def roll(*args, **kwargs):
```
twiecki: all these need doc-strings.

empyrical/utils.py
```diff
@@ -24,12 +26,63 @@
...
+
+def _roll_ndarray(func, window, *args, **kwargs):
+    data = []
+    for i in range(window, len(args[0])):
+        rets = [s[i-window:i] for s in args]
+        data.append(func(*rets, **kwargs))
+    return np.array(data)
+
+
+def _roll_pandas(func, window, *args, **kwargs):
```
twiecki: these helper functions definitely need tests, easy for edge cases / errors to creep in.

empyrical/stats.py
```diff
@@ -1005,6 +993,214 @@ def cagr(returns, period=DAILY, annualization=None):
...
+
+    Parameters
+    ----------
+    see documentation for `capture` (pass all args, kwargs required)
+
+    window : int, required
+        Size of the rolling window in terms of the periodicity of the data.
+        - eg window = 60, periodicity=DAILY, represents a rolling 60 day window
+    """
+    return roll(returns, factor_returns, window=window, functions=[
```
twiecki: couldn't you pass `functions=[up_capture]` instead? If functions is ever only a single argument might make the code below easier too. Or a lambda could be used here, or something neat from functools or itertools.

twiecki:
@cgdeboer This looks great, many thanks for the refactor / contribution. I left some comments, mainly tests for the utils functions would be good. But generally I find the way you coded things up very elegant. Let me know if you have any questions.

cgdeboer:
@twiecki I'll address your comments by the end of this week. Thanks for the look !

setup.py
cgdeboer: From `try:except` block in the `utils.py` file it looks like the use of the bottleneck library is intended to be optional (for those who need the bone rattlin' speed of optimized C ).  Removing bottleneck from the hard requirements is merely a nod to the logic in the `utils.py` file. 

It seems like we might want to either: 

1. remove bottleneck from the requirements (as I have done), or 
2. simplify the utils file to always use the bottleneck library. Either seem logical to me. 

@twiecki 's choice 😃 

(if there is some other reason to keep it as it originally was, I'm happy to revert to that as well... perhaps some sort of thing where `bottleneck` is only supported in certain architectures? )

richafrank: I think bottleneck needs to be required at the moment. We [discovered](https://github.com/quantopian/empyrical/pull/45#issuecomment-299338946) that it has diverged slightly from numpy, and we rely on the bottleneck behavior.

empyrical/stats.py
cgdeboer: Added em in a89510e . I also realized I added capture to the wrong list, it belongs to FACTOR_STAT_FUNCS, so I put capture (reg, up and down) all in there.

empyrical/stats.py
cgdeboer: Added in a89510e

empyrical/stats.py
cgdeboer: Not sure how that happened (the tab in there). Also, I don't think I intended to keep the doc string "Returns" syntax with the `list` nomenclature, that was some WIP I had in there to remind me that a list came back with that. I've reverted this to the original "pair o' floats" style "Returns" section, and ensured that's how it is across the board.

setup.py
cgdeboer: noted.

cgdeboer: Fixed in aa07b5c

empyrical/utils.py
cgdeboer: Fixed in 6049309

empyrical/utils.py
cgdeboer: added in 	ab1eabe

empyrical/stats.py
cgdeboer: I think your idea about a single function makes the most sense. The `functions=[]` was sort of over-engineered. I changed it all to take a `function` kwarg in 	71ea444

cgdeboer:
@twiecki I think all of your comments have been addressed. Looks like ole Travis is just finishing up the CI testing (or maybe hung up on one build job).

empyrical/stats.py
```diff
@@ -1005,6 +992,227 @@ def cagr(returns, period=DAILY, annualization=None):
...
+
+    Parameters
+    ----------
+    see documentation for `capture` (pass all args, kwargs required)
+
+    window : int, required
+        Size of the rolling window in terms of the periodicity of the data.
+        - eg window = 60, periodicity=DAILY, represents a rolling 60 day window
+    """
+    return roll(returns, factor_returns, window=window, functions=up_capture,
```
twiecki: function

cgdeboer: Newman !! Fixed it in 6efeee1

cgdeboer:
Alrighty @twiecki , addressed that last pluralized one. Let me know if you see anything else, could catch.

twiecki:
Did the unittest catch it as well?

cgdeboer:
Well. You know what, I don't think it did. Perhaps I need to write a functional test for each of the new methods. I sort of thought we'd rely on the individual stat tests for "financial" accuracy, and a sample of tests for the wrappers like `up`, `down`, and `roll`.  

I could see it going either way. What is your preference?

twiecki:
Best to have it run through all code paths.

cgdeboer:
Will add tests for all the methods. Back in a few.

twiecki:
@cgdeboer Any update?

cgdeboer:
@twiecki should have em all done this week. (My few referenced above turned from a few hours to a few days). Other stuff hopped in the way. Shooting for Thursday morning.

twiecki:
OK, no rush obviously, just wondered if we could help.

cgdeboer:
Most of the new tests are now written, just taking some extra time to double check the inputs vs expected.

cgdeboer:
@twiecki tests added. 🏀  is in your court.

empyrical/utils.py
```diff
@@ -12,24 +12,125 @@
...
-    # slower numpy
-    import numpy as np
-    nanmean = np.nanmean
-    nanstd = np.nanstd
-    nansum = np.nansum
-    nanmax = np.nanmax
-    nanmin = np.nanmin
-    nanargmax = np.nanargmax
-    nanargmin = np.nanargmin
+nanmean = bn.nanmean
```
twiecki: This should be reverted.

cgdeboer: Reverted in dfe495f.

twiecki:
Great work, thanks @cgdeboer!

cgdeboer:
Thanks !. @twiecki  I'll keep an eye on https://pypi.python.org/pypi/empyrical/ for version update ?

twiecki:
@cgdeboer New release should be up.

PR #48, opened 2017-05-30
Add VaR and CVaR functions

dmichalowicz:
Add a function that calculates the [conditional value at risk](https://en.wikipedia.org/wiki/Expected_shortfall) of a set of returns streams.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -1214,6 +1214,55 @@ def roll_sharpe_ratio(returns, window=10, **kwargs):
     return roll(returns, window=window, function=sharpe_ratio, **kwargs)
 
 
+def value_at_risk(returns, cutoff=0.05):
+    """
+    Value at risk (VaR) of a returns stream.
+
+    Parameters
+    ----------
+    returns : pandas.Series or 1-D numpy.array
+        Non-cumulative daily returns.
+    cutoff : float, optional
+        Decimal representing the percentage cutoff for the bottom percentile of
+        returns. Defaults to 0.05.
+
+    Returns
+    -------
+    VaR : float
+        The VaR value.
+    """
+    return np.percentile(returns, 100 * cutoff)
+
+
+def conditional_value_at_risk(returns, cutoff=0.05):
+    """
+    Conditional value at risk (CVaR) of a returns stream.
+
+    CVaR measures the expected single-day returns of an asset on that asset's
+    worst performing days, where "worst-performing" is defined as falling below
+    ``cutoff`` as a percentile of all daily returns.
+
+    Parameters
+    ----------
+    returns : pandas.Series or 1-D numpy.array
+        Non-cumulative daily returns.
+    cutoff : float, optional
+        Decimal representing the percentage cutoff for the bottom percentile of
+        returns. Defaults to 0.05.
+
+    Returns
+    -------
+    CVaR : float
+        The CVaR value.
+    """
+    # PERF: Instead of using the 'value_at_risk' function to find the cutoff
+    # value, which requires a call to numpy.percentile, determine the cutoff
+    # index manually and partition out the lowest returns values. The value at
+    # the cutoff index should be included in the partition.
+    cutoff_index = int((len(returns) - 1) * cutoff)
+    return np.mean(np.partition(returns, cutoff_index)[:cutoff_index + 1])
+
+
 SIMPLE_STAT_FUNCS = [
     cum_returns_final,
     annual_return,
@@ -1227,7 +1276,9 @@ SIMPLE_STAT_FUNCS = [
     stats.skew,
     stats.kurtosis,
     tail_ratio,
-    cagr
+    cagr,
+    value_at_risk,
+    conditional_value_at_risk,
 ]
 
 FACTOR_STAT_FUNCS = [
--- a/empyrical/tests/test_stats.py
+++ b/empyrical/tests/test_stats.py
@@ -1197,6 +1197,54 @@ class TestStats(TestCase):
             expected,
             DECIMAL_PLACES)
 
+    def test_value_at_risk(self):
+        value_at_risk = self.empyrical.value_at_risk
+
+        returns = [1.0, 2.0]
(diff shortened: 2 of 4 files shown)


empyrical/stats.py
```diff
@@ -1005,6 +1005,41 @@ def cagr(returns, period=DAILY, annualization=None):
     return ending_value ** (1. / no_years) - 1
 
 
+def value_at_risk(returns, cutoff):
```
ssanderson: If this gets passed a 2-d array (which would match the expected input for `conditional_value_at_risk`, then `np.percentile` will calculate the percentile of the flattened array:

```
In [5]: data
Out[5]:
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19],
       [20, 21, 22, 23, 24]])

In [6]: np.percentile(data, 30)
Out[6]: 7.1999999999999993
```

As a user, I would expect `value_at_risk` and `conditional_value_at_risk` to have the same input/output signature.

empyrical/stats.py
```diff
@@ -1005,6 +1005,41 @@ def cagr(returns, period=DAILY, annualization=None):
...
+    return np.percentile(returns, 100 * cutoff)
+
+
+def conditional_value_at_risk(returns,
+                              weights,
+                              cutoff=0.05,
+                              lookback_days=None):
```
ssanderson: As a user, when would I want to pass `lookback_days` to this vs. just slicing the dataframe myself?

dmichalowicz:
@ssanderson I've updated this based on our recent discussions, mind taking a look?

empyrical/stats.py
```diff
@@ -1005,6 +1005,46 @@ def cagr(returns, period=DAILY, annualization=None):
...
+
+    Returns
+    -------
+    CVaR : float
+        The CVaR value.
+    """
+    # PERF: Instead of using the 'value_at_risk' function to find the cutoff
+    # value, which requires a call to numpy.percentile, determine the cutoff
+    # index manually and partition out the lowest returns values.
+    num_low_days = int((len(returns) - 1) * cutoff) + 1
```
ssanderson: It feels like there should be a simpler way to take this slice without having to add and subtract 1 three times.  What's the intended behavior of all the shuffling here?

empyrical/stats.py
```diff
@@ -1005,6 +1005,46 @@ def cagr(returns, period=DAILY, annualization=None):
...
+    return np.mean(np.partition(returns, num_low_days - 1)[:num_low_days])
+
+
```
ssanderson: The implementations of these functions look correct to me.  We should probably update the docstrings to explain what these functions are actually for and where/when you would want to use them.  For example, for cvar, we might want something like:

```python
def conditional_value_at_risk(returns, cutoff=0.05):
    """
    Conditional value at risk (CVaR) of a returns stream.

    CVaR measures the expected single-day returns of an asset on that asset's
    worst-performing day, where "worst-performing" is defined as falling below
    ``cutoff`` as a percentile of all daily returns.

    Parameters
    ----------
    ...
    """
```

cc @marketneutral @CaptainKanuk @jmccorriston for thoughts on language here.

empyrical/tests/test_stats.py
```diff
@@ -989,6 +989,33 @@ def test_cagr_noisy(self, returns, add_noise):
...
+            conditional_value_at_risk(returns, cutoff=1), expected_cvar,
+        )
+
+        # Test a returns stream of 21 data points at different cutoffs.
+        returns = np.random.normal(0, 0.02, 21)
+
+        for cutoff in (0, 0.0499, 0.05, 0.20, 0.999, 1):
+            # Find the VaR based on our cutoff, then take the average of all
+            # values at or below the VaR.
+            var = value_at_risk(returns, cutoff)
```
ssanderson: Seems like we'd weant a test for `value_at_risk` by itself if we're using that as the definition of correctness for `conditional_value_at_risk`?

dmichalowicz: I had originally omitted a test as it felt like I was simply testing numpy's `percentile` function; but it's probably worth having one in case we ever change `value_at_risk`, so I'll add

empyrical/stats.py
dmichalowicz: The only simpler way I can think of is to rename `num_low_days` to something like `index_of_cutoff`. That way I don't have to do `+1` here, or `-1` below. Does that seem better/less confusing?

empyrical/tests/test_stats.py
dmichalowicz: added

empyrical/stats.py
```diff
@@ -1214,6 +1214,55 @@ def roll_sharpe_ratio(returns, window=10, **kwargs):
...
+    """
+    # PERF: Instead of using the 'value_at_risk' function to find the cutoff
+    # value, which requires a call to numpy.percentile, determine the cutoff
+    # index manually and partition out the lowest returns values. The value at
+    # the cutoff index should be included in the partition.
+    cutoff_index = int((len(returns) - 1) * cutoff)
+    return np.mean(np.partition(returns, cutoff_index)[:cutoff_index + 1])
+
+
 SIMPLE_STAT_FUNCS = [
```
twiecki: please add them to the list below.

ssanderson:
LGTM pending https://github.com/quantopian/empyrical/pull/48#discussion_r123076152

setup.py
```diff
@@ -19,7 +19,8 @@
 
 
 DISTNAME = "empyrical"
-DESCRIPTION = "empyrical is a Python library with performance and risk statistics commonly used in quantitative finance"
+DESCRIPTION = """empyrical is a Python library with performance and risk \
+statistics commonly used in quantitative finance"""
```
dmichalowicz: @twiecki I changed this back (with an added backslash) due to flake8 failing.

twiecki: Thanks, is this a single line then?

dmichalowicz: yup!

PR #49, opened 2017-06-12
Fix sortino_ratio downside_risk

lostAstronaut:
This is needed because if yearly data is given, then the Sortino ratio
(sortino_ratio) can be overridden by setting the annualization value
but when downside risk is calculated in the downside_risk function it
will default to daily as annualization is not carried forward thus
affecting the final value.


twiecki:
Thanks @lostAstronaut!

twiecki:
@lostAstronaut Sorry, I had to revert, we need to fix the unittests and reopen.

lostAstronaut:
I agree, should I reopen?

twiecki:
Yes, please.

lostAstronaut:
Done, cheers

Issue #59, opened 2017-10-04
Bug in annualization of alpha?


twiecki:
https://github.com/quantopian/empyrical/blob/master/empyrical/stats.py#L821 multiplies daily alpha by the annualization factor (252). However, as daily alphas compound just like returns, the correct way to annualize is `(daily_alpha + 1) ** ann_factor - 1`.

PR #60, opened 2017-10-16
BUG: fix annualization of alpha

vikram-narayan:
For #59, annualize alpha correctly


twiecki:
This looks right. But would hope that some tests catch this we need to change.

vikram-narayan:
@twiecki just updated `test_stats.py` to fix the failures - mind taking a quick look at those as well? Mostly just updating numbers

empyrical/tests/test_stats.py
```diff
@@ -1111,7 +1114,7 @@ def test_roll_up_capture(self, returns, factor_returns, window, expected):
         (empty_returns, simple_benchmark, (np.nan, np.nan)),
         (one_return, one_return, (np.nan, np.nan)),
         (mixed_returns[1:], negative_returns[1:],
-         (-8.306666666666668, -0.71296296296296313)),
+         (-0.9997853834885004, -0.71296296296296313)),
```
twiecki: This already shows why the previous result was wrong.

twiecki:
LGTM. We should probably alert others to this bugfix.

empyrical/stats.py
```diff
@@ -818,7 +818,7 @@ def alpha_aligned(returns, factor_returns, risk_free=0.0, period=DAILY,
     adj_factor_returns = _adjust_returns(factor_returns, risk_free)
     alpha_series = adj_returns - (_beta * adj_factor_returns)
 
-    return nanmean(alpha_series) * ann_factor
+    return (nanmean(alpha_series) + 1) ** ann_factor - 1
```
yankees714: What would you think about breaking this formula out into a helper function? Feels like that would make it clearer what's happening here, and less error prone.

(At the least a comment to the effect of #59 would be good)

vikram-narayan: added an `annualize` helper in utils, and used that here

empyrical/utils.py
```diff
@@ -443,3 +443,10 @@ def default_returns_func(symbol, start=None, end=None):
         rets = get_symbol_returns_from_yahoo(symbol, start=start, end=end)
 
     return rets[symbol]
+
+
+def annualize(stat, ann_factor=252):
```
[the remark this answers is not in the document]
chrismyles4: Nice catch, I agree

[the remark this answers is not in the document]
vikram-narayan: 👍  renamed to `annualize_compounding`

chrismyles4:
FYI I have walked Vikram's analysis of where this code is used to the PM team, Dan and Rich. We are good to go, they will be adding a forum post to explain the change.

> I agree that a forum announcement is appropriate and will be a non issue.  "we love open source software, our calculations are all out there, if you see any mistakes we're all better off when we fix them."

yankees714:
LGTM

twiecki:
Can we merge this?

vikram-narayan:
We chose to wait for when we have the bandwidth to write a blog post announcing the change. Will follow up on that today

empyrical/stats.py
```diff
@@ -1279,6 +1280,9 @@ def alpha_aligned(returns,
     adj_factor_returns = _adjust_returns(factor_returns, risk_free)
     alpha_series = adj_returns - (_beta * adj_factor_returns)
 
+<<<<<<< HEAD
```
twiecki: conflict

twiecki:
I think there is still a bug here, as the failing tests show. Seems like it always returns 0 or nan?

twiecki:
Thanks @gmanoim-quantopian !

PR #63, opened 2017-11-10
MAINT: strip whitespace from get_fama_french

vikram-narayan:


empyrical/utils.py
```diff
@@ -234,6 +234,9 @@ def get_fama_french():
     five_factors /= 100.
     five_factors.index = five_factors.index.tz_localize('utc')
 
+    five_factors.columns = [col.strip() for col in
+                            five_factors.columns]
```
yankees714: I think you could do this without the list comprehension as `five_factors.columns.str.strip()`, totally your call on which you prefer

vikram-narayan: switched to this

PR #66, opened 2017-11-26
Removed ellipsis in Pandas Support

riabaldevia:
Defining returns rendered a TypeError: unsupported operand type(s) for -: 'ellipsis' and 'float'.


twiecki:
Thanks @riabaldevia. Not sure what that was even doing there.

PR #67, opened 2017-11-27
Travis pandas numpy wheels

richafrank:
Goal is to fix the build on master which is currently failing py2.7 with:
```python
ImportError: C extension: numpy.core.multiarray failed to import not built. If you want to import pandas from the source directory, you may need to run 'python setup.py build_ext --inplace --force' to build the C extensions first.
```
because the pandas wheel was compiled against a different version of numpy than we're installing.


twiecki:
LGTM, thanks @richafrank!

richafrank:
@twiecki My changes to the build matrix ok with you?

twiecki:
Yes!

PR #62, opened 2017-11-10
Fixed #61, annualization of Sortino Ratio

gabrieleiacono:

Issue #61, opened 2017-11-09
Wrong annualization of Sortino Ratio?


gabrieleiacono:
The Sortino Ratio is annualized in stats.py as follows:
`return sortino * ann_factor`

where `ann_factor` is
`ann_factor = annualization_factor(period, annualization)`

I believe you actually meant to return
`return sortino * np.sqrt(ann_factor)`

just like you did with the Sharpe Ratio.

twiecki:
Good catch, want to do a PR?

rsheftel:
The math is actually correct in the existing version. The confusion is because of where the `ann_factor` multiplication occurs. The reason it is correct is because the calculation of the downside risk with the `downside_risk` function returns an annualized value, it does the `* np.sqrt()` in that function. The calculation of the numerator of returns in the `sortino_ratio()` function does not annualize:

`mu = nanmean(adj_returns, axis=0)` so the code annualizes at the end at the return value.
`return sortino * ann_factor'

Agree this is confusing, so I submitted a pull request that just moves the `* ann_factor` to the line that calculates the returns so the math is more obvious:

`mu = nanmean(adj_returns, axis=0) * ann_factor`

twiecki:
@rsheftel Thanks, very glad you point this out before we merged it.

twiecki:
Closed by #74.

gabrieleiacono:
I'm terribly sorry, thank you @riabaldevia for finding my mistake. It is much better now, after PR #74.

PR #74, opened 2018-01-26
Proper change to Sortino ratio to clarify the math

rsheftel:
move the annualization factor multiplication to make it more obvious that the math is correct.

This is a cosmetic change to where the annualization of the numerator returns is located. This seems to have caused confusion, see issue #61 

This PR should also replace #62 which is incorrect and should be accepted.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -521,8 +521,9 @@ def sortino_ratio(returns, required_return=0, period=DAILY,
     average_annual_return = nanmean(adj_returns, axis=0) * ann_factor
     annualized_downside_risk = (_downside_risk if _downside_risk is not None
                                 else downside_risk(returns, required_return,
-                                            period=period,
-                                            annualization=annualization))
+                                                   period=period,
+                                                   annualization=
+                                                   annualization))
     sortino = average_annual_return / annualized_downside_risk
     return sortino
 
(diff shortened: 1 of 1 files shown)


empyrical/stats.py
```diff
@@ -515,12 +515,12 @@ def sortino_ratio(returns, required_return=0, period=DAILY,
     ann_factor = annualization_factor(period, annualization)
 
     adj_returns = _adjust_returns(returns, required_return)
-    mu = nanmean(adj_returns, axis=0)
+    mu = nanmean(adj_returns, axis=0) * ann_factor
```
twiecki: maybe to be extra-clear call this `mean_returns_ann`

empyrical/stats.py
```diff
@@ -515,12 +515,12 @@ def sortino_ratio(returns, required_return=0, period=DAILY,
...
     dsr = (_downside_risk if _downside_risk is not None
```
twiecki: dsr_ann

rsheftel:
Made the function variables more explanatory and added a link to a good paper to describes the math and why many implementations are not mathematically correct (but the empyrical one is correct, so we are all good there)

rsheftel:
Sorry for all the commits to fix the flake8 check but everything passes now and is ready to merge if you are happy with the code.

twiecki:
Thanks @rsheftel!

PR #22, opened 2016-08-24
Add docs

ahgnaw:
Use sphinx to generate documentation pages in a separate directory 'empyrical-docs'.

PR #75, opened 2018-01-30
ReadTheDoc setup for auto-generation of documentation

rsheftel:
This PR does the following:

- Creates a new directory "docs" for the documentation items
- Populates the docs directory with the items required for automatic documentation generation using sphinx. If the repository is connected to ReadTheDocs then ReadTheDocs will automatically generate the documentation on each commit.
- Makes a small modification to setup.py to work with ReadTheDocs
- Cleaned up the docstrings in stats and utils modules to properly auto-generate the documentation. None of the actual code was changed, just the docstrings
- I also bumped up the min version of pandas to 0.19 because that is what pandas_datareader requires itself anyway.

For an example of how ReadTheDocs renders the documentation I connected my fork to RTD. The docs website is here: http://rsheftel-empyrical.readthedocs.io/en/latest/index.html
(Note I left the project name "empyrical" free so that the official documentation can use that URL)

Note that to have RTD build the docs correctly you will need to make sure the following setting in the advanced are set:

- Install Project is checked
- Python configuration file is "docs/conf.py"
- "Use system packages: Get the virtual environment…" is checked
- Python Interpreter is "Cpython 3.x"

And that is it. If the webhooks are set properly then every commit will regenerate the docs automatically.

If you need help with setting up RTD let me know. I believe you can add me as an administrator and I can help get it to work.

This replaces the pull request #22


setup.py
```diff
@@ -59,12 +59,17 @@
 
 requirements = [
     'numpy>=1.9.2',
-    'pandas>=0.16.1',
```
twiecki: I think we need to stay on this old version for now because of zipline.

twiecki:
Wow, thank you so much! Especially fixing the doc-strings is extremely helpful.

docs/conf.py
```diff
@@ -0,0 +1,155 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = 'empyrical'
+copyright = '2016, Quontopian'
```
twiecki: 2018

twiecki: Quantopian

docs/empyrical.rst
```diff
@@ -0,0 +1,45 @@
+empyrical package
+=================
+
+Subpackages
```
twiecki: these don't seem to add anything in the docs

docs/empyrical.rst
```diff
@@ -0,0 +1,45 @@
...
+-----------
+
+.. toctree::
+
+    empyrical.tests
+
+Submodules
+----------
+
+empyrical\.periods module
```
twiecki: don't need this

docs/empyrical.tests.rst
```diff
@@ -0,0 +1,22 @@
+empyrical\.tests package
```
twiecki: probably don't need this either

rsheftel: Will delete this file in the next commit, but again it will re-appear if make_docs.bat is run again

rsheftel:
@twiecki all suggested changes have been made and all tests pass.  If you decide to merge this and setup the docs on RTD, then you should change the link on the title of the empyrical to point to the RTD docs and not the GitHub pages. (I don't have access to change that)

twiecki:
Thanks @rsheftel!

PR #76, opened 2018-01-31
MAINT: add perf attrib to empyrical

vikram-narayan:
- move core perf attrib computation logic and tests from pyfolio to empyrical


empyrical/__init__.py
```diff
@@ -62,3 +63,10 @@
     MONTHLY,
     YEARLY
 )
+
+
+from .perf_attrib import (
+    perf_attrib,
+    compute_exposures,
+    create_perf_attrib_stats,
+)
```
ssanderson: Missing a newline here.  You probably can configure your editor to make it always do this automatically for files with .py suffixes.

richafrank: Shall we add flake8 to the build?

empyrical/perf_attrib.py
```diff
@@ -0,0 +1,409 @@
...
+                                        factor_returns,
+                                        factor_loadings,
+                                        transactions,
+                                        pos_in_dollars=pos_in_dollars)
+
+    # Note that we convert positions to percentages *after* the checks
+    # above, since get_turnover() expects positions in dollars.
+    positions = _stack_positions(positions, pos_in_dollars=pos_in_dollars)
+
+    risk_exposures_portfolio = compute_exposures(positions,
```
ssanderson: I think the only logic we actually need for perf attribution in zipline is the code from here through the end of this function.  In general, my mental model is that, as much as possible, empyrical should be just the "bare bones" calculations.  What do you think about carving out just the logic of "Given returns, positions, factor_returns, and factor_loadings, compute perf attribution" and leave the logic of "display warnings about missing/unreasonable data" to pyfolio?

empyrical/perf_attrib.py
```diff
@@ -0,0 +1,409 @@
...
+            ('Annualized Return', annualized_returns_by_factor),
+            ('Cumulative Return', cumulative_returns_by_factor),
+        ]),
+        index=risk_exposures.columns,
+    )
+
+    return summary, risk_exposure_summary
+
+
+def _align_and_warn(returns,
```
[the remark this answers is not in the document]
vikram-narayan: I moved this since it includes logic to make the indexes match up in addition to the warnings stuff. Should I make it so the warnings only get displayed if it's called from pyfolio?

empyrical/__init__.py
vikram-narayan: fixed. it looks like it's already in the build? not sure why it missed the newline:
```
before_script:
  - "flake8 ."
```

empyrical/perf_attrib.py
vikram-narayan: I can do that. Are we assuming that anything calling into empyrical has its inputs already correctly formatted?

vikram-narayan: only concern is that it'll result in duplicating some of the logic to make indexes match up, etc., before perf attrib gets called

empyrical/perf_attrib.py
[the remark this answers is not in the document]
ssanderson: That's what I was thinking.  I see the warnings as being part of the tearsheet UX, not as part of the calculations themselves.

empyrical/perf_attrib.py
ssanderson: I think so.  Otherwise we're emposing a performance penalty on a caller that already has well-aligned data.  In PyFolio, the expected user is running a tearsheet interactively in a notebook, so the extra overhead of doing data validation isn't a big deal, but I don't think we want to make the assumption that those checks are always an acceptable overhead.

ssanderson: Also, in zipline, there are cases where we won't have all the transactions in memory to use for turnover (e.g., in live algorithms, we don't want to reload all the historical transactions).

vikram-narayan:
empyrical updated to just include perf_attrib and compute_exposures, moving the rest back to pyfolio

empyrical/perf_attrib.py
```diff
@@ -0,0 +1,130 @@
+import pandas as pd
+
+
+def perf_attrib(returns,
+                positions,
+                factor_returns,
+                factor_loadings):
+    """
+    Does performance attribution given risk info.
```
ssanderson: It would be nice to have a little more information in this docstring about what performance attribution is and what it's for.  Just saying `perf_attrib` is a function that "does performance attribution" doesn't really provide much information to a reader.

vikram-narayan: added a more detailed docstring

empyrical/perf_attrib.py
```diff
@@ -0,0 +1,141 @@
...
+                factor_loadings):
+    """
+    Attributes the performance of a returns stream to a set of risk factors.
+
+    Performance attribution determines how much each risk factor, e.g.,
+    momentum, the technology sector, etc., contributed to total returns, as
+    well as the daily exposure to each of the risk factors. The returns that
+    can be attributed to one of the given risk factors are the
+    `common_returns`, and the returns that _cannot_ be attributed to a risk
+    factor are the `specific_returns`, or the alpha. The common_returns and
```
ssanderson: My instinct would probably be to leave out the reference to alpha here.  I get the impression that that's kind of a loaded term, and that we use "specific vs. common returns" in part to avoid confusion around what different people expect/mean by "alpha".

vikram-narayan: Yup makes sense. My thought in including it was that people might look for "alpha" when poking around empyrical, and might gloss over specific/common returns. But I can remove it if it obscures rather than clarifies what perf attrib is meant for

ssanderson: cc @jzhang18 for thoughts on this.

vikram-narayan: changed "or the alpha" to be "or the alpha returns" on @jzhang18's suggestion

vikram-narayan: discussed a little more, the reference to alpha is removed

empyrical/perf_attrib.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def compute_exposures(positions, factor_loadings):
+    """
+    Compute daily risk factor exposures.
+
+    Parameters
+    ----------
+    positions: pd.Series
+        A series of holdings indexed by date and ticker.
```
ssanderson: Can we note here that the expectation is that positions are percentages?

vikram-narayan: added `as percentages` here

PR #78, opened 2018-02-01
Fix duplicate functions: annual_return and cagr

rsheftel:
In the stats.py there are two functions that do the identical math: annual_return() and cagr(). This PR does the following:

    Move the two functions next to each-other so it is obvious they are the same
    Have the function cagr() just call annual_return()
    Clean up the code for annual_return() so that it is more readable

Also I cleaned up a small style point on the annual_volatility() function to be consistent with the annual_return() style. Previously the function assigned the result to a variable and then return variable, now the math is right in the return line.

This PR replaces #77


twiecki:
Very nice, thanks!

PR #83, opened 2018-02-09
ENH: support rolling calculations through vectorization

llllllllll:
This also adds an `out` argument for many functions so that you can pre-allocate the output array.

This is useful for speeding up `alpha_beta` even further by preallocating the 2d array to write the output, and then passing `alpha` and `beta` the column slices to fill. You an also use this with `roll_*` to efficiently pre-fill the window with nan or 0, for example:

```python
window = 126
beta_to_spy = np.empty(len(returns), np.nan)
beta_to_spy[:window] = np.nan
empyrical.roll_beta_aligned(returns, spy_returns, out=beta_to_spy[window:])
```


empyrical/utils.py
```diff
@@ -447,3 +463,72 @@ def default_returns_func(symbol, start=None, end=None):
...
+           [[10, 11, 12, 13, 14],
+            [15, 16, 17, 18, 19]],
+    <BLANKLINE>
+           [[15, 16, 17, 18, 19],
+            [20, 21, 22, 23, 24]]])
+    """
+    orig_shape = array.shape
+    if not orig_shape:
+        raise IndexError("Can't restride a scalar.")
+    elif orig_shape[0] < length:
```
prsutherland: Might want to also check the `length == 0` case too

prsutherland: Also when length is a float, there's strange behavior. Using `a` from the docstring:

```
>>> rolling_window(a, 2.01)
array([[[ 0,  1,  2,  3,  4],
        [ 5,  6,  7,  8,  9]],

       [[ 5,  6,  7,  8,  9],
        [10, 11, 12, 13, 14]],

       [[10, 11, 12, 13, 14],
        [15, 16, 17, 18, 19]]])
```

empyrical/stats.py
```diff
@@ -20,11 +20,102 @@
...
+        -------
+        rolling_{name} : array-like
+            The rolling {human_readable}.
+        """
+        allocated_output = out is None
+
+        if len(arr):
+            out = function(rolling_window(arr, window).T, out=out, **kwargs)
+        else:
+            out = np.empty(0, dtype='float64')
```
prsutherland: I'm not sure what typical behavior is, but if the user passes `out`, do we need to zero the `out` array?

empyrical/stats.py
```diff
@@ -20,11 +20,102 @@
...
+            Forwarded to :func:`~empyrical.{name}`.
+
+        Returns
+        -------
+        rolling_{name} : array-like
+            The rolling {human_readable}.
+        """
+        allocated_output = out is None
+
+        if len(arr):
```
prsutherland: Should we check that `len(arr) >= window` or should we just let `rolling_window` fail loudly?

empyrical/stats.py
```diff
@@ -20,11 +20,102 @@
...
+        Returns
+        -------
+        rolling_{name} : array-like
+            The rolling {human_readable}.
+        """
+        allocated_output = out is None
+
+        if window >= 1 and len(lhs) and len(rhs):
+            out = function(
+                rolling_window(lhs, window).T,
```
prsutherland: I'm a bit confused why the rolling window is being transposed.

empyrical/utils.py
llllllllll: what needs to be done in the `length == 0` case? You will get an array with a dimension of size 0, but that is fine.

Also, I think the second thing is fine, we say that it takes an `int`, so I wouldn't expect it to work with non-int fields.

empyrical/stats.py
llllllllll: This isn't zeroing the array, this is creating a length 0 array.

empyrical/stats.py
llllllllll: Because the functions allow dataframes, axis 0 is the time dimension and axis 1 are the unique time series. Rolling window adds the new dimension to the front, so we need to transpose.

empyrical/utils.py
```diff
@@ -447,3 +463,72 @@ def default_returns_func(symbol, start=None, end=None):
...
+    Parameters
+    ----------
+    array : np.ndarray
+        The base array.
+    length : int
+        Length of the synthetic first axis to generate.
+
+    Returns
+    -------
+    out : np.ndarray
```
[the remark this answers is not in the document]
llllllllll: This has much less risk than `as_strided` directly because we make assertions about the new shape. The reason `as_strided` is dangerous is that you can use it to read arbitrary memory. Mutating a value in the result will mutate all occurrences of the value across all the time-slices, but given that we say, "restride an array" in the docstring, I think that isn't surprising.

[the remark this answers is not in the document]
ssanderson: I agree with @prsutherland that a warning against mutation is probably appropriate here.  I don't think the expected user of empyrical is going to know offhand that "restrided array" means "playing fancy tricks with numpy to re-use the same memory location for multiple cells in the array".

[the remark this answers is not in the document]
llllllllll: Why not just make it immutable then?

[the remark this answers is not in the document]
ssanderson: I'm fine with making this return a read-only array by default.  Then the aforementioned warning can go on the parameter description for the `readonly` flag, which I think should have a default of True.

empyrical/stats.py
llllllllll: rolling_window gives a reasonable error here.

PR #85, opened 2018-02-27
BUG: Preserve output index in `cum_returns`.

ssanderson:
This also now makes all our rolling operations return an output indexed by a suffix of the input index.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -37,7 +37,8 @@ def _create_unary_vectorized_roll_function(function):
         window : int
             Size of the rolling window in terms of the periodicity of the data.
         out : array-like, optional
-            The array to store the store the output.
+            Array to use as output buffer.
+            If not passed, a new array will be created.
         **kwargs
             Forwarded to :func:`~empyrical.{name}`.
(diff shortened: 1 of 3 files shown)


empyrical/stats.py
```diff
@@ -37,7 +37,8 @@ def unary_vectorized_roll(arr, window, out=None, **kwargs):
         window : int
             Size of the rolling window in terms of the periodicity of the data.
         out : array-like, optional
-            The array to store the store the output.
```
llllllllll: This typo...

empyrical/tests/test_stats.py
```diff
@@ -1455,6 +1492,9 @@ class Test2DStatsArrays(Test2DStats):
     def empyrical(self):
         return PassArraysEmpyricalProxy(self, np.ndarray)
 
+    def assert_indexes_match(self, result, expected):
```
llllllllll: cute

llllllllll:
lgtm

twiecki:
We should probably cut a new bugfix release.

PR #86, opened 2018-03-12
BUG: Preserve columns in cum_returns.

ssanderson:
- Fixes a bug where we failed to maintain columns when calling
  `empyrical.cum_returns` on a DataFrame.


ssanderson:
It looks to me like https://github.com/quantopian/pyfolio/blob/master/pyfolio/tests/test_perf_attrib.py#L176-L180 should have caught this?

PR #92, opened 2018-04-17
MAINT: use int for groupby

vikram-narayan:
Instead of using the index name in the call to groupby, use the index level number.

### DIFF
--- a/empyrical/perf_attrib.py
+++ b/empyrical/perf_attrib.py
@@ -138,4 +138,4 @@ def compute_exposures(positions, factor_loadings):
             2017-01-02  0.821872  1.520515
     """
     risk_exposures = factor_loadings.multiply(positions, axis='rows')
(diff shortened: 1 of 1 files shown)


empyrical/perf_attrib.py
```diff
@@ -138,4 +138,4 @@ def compute_exposures(positions, factor_loadings):
             2017-01-02  0.821872  1.520515
     """
     risk_exposures = factor_loadings.multiply(positions, axis='rows')
-    return risk_exposures.groupby(level='dt').sum()
```
ssanderson: Is there a test we can add/modify to go along with this change?

ssanderson: LGTM assuming tests pass

ssanderson:
LGTM assuming tests pass

PR #93, opened 2018-04-24
TST Add test for inf Sharpe

gusgordon:


richafrank:
Looks like a nice addition to me, so we can know what to expect. Any thoughts @llllllllll ?

llllllllll:
Is inf the desired value? If so, this seems good.

gusgordon:
Yeah - spoke with Thomas about it and he was also OK with it being inf. The nan value is unreliable anyways due to float error, eg `ep.sharpe_ratio([1, 1, 1, 1, 1])` is inf but `ep.sharpe_ratio([0.1, 0.1, 0.1, 0.1, 0.1])` is a large number.

PR #94, opened 2018-05-01
fix beta(series, dataframe)

llllllllll:
also, remove uses of unseeded random in the test suite.

### DIFF
--- a/empyrical/stats.py
+++ b/empyrical/stats.py
@@ -906,6 +906,32 @@ def excess_sharpe(returns, factor_returns, out=None):
 roll_excess_sharpe = _create_binary_vectorized_roll_function(excess_sharpe)
 
 
+def _to_pandas(ob):
+    """Convert an array-like to a pandas object.
+
+    Parameters
+    ----------
+    ob : array-like
+        The object to convert.
+
+    Returns
+    -------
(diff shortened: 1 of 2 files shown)


empyrical/stats.py
```diff
@@ -934,7 +960,7 @@ def _aligned_series(*many_series):
     # dataframe has no ``itervalues``
     return (
         v
-        for _, v in iteritems(pd.concat(map(pd.Series, many_series), axis=1))
+        for _, v in iteritems(pd.concat(map(_to_pandas, many_series), axis=1))
```
ehebert: Should we add a test in `test_stats` to exercise the `DataFrame` conversion. Or modify one of the existing tests such as the `test_beta`, to pass both a `Series` and a `DataFrame`?

llllllllll: The new test I added to `test_beta` passes a series as the lhs and a dataframe as the rhs

ehebert: Ah, I had just been looking at this commit. I see it now, https://github.com/quantopian/empyrical/pull/94/commits/660af2d443e39278eff459d76a4154dccba4f8d0#diff-20db41cf8e5a7df30bbccfa9afb81f44R880

PR #89, opened 2018-03-26
 Change default_returns_func to use Quandl API

jaycode:
I updated `default_returns_func` to use Quandl, created `get_symbol_returns_from_quandl`, and deprecated `get_symbol_returns_from_yahoo`.

PR #96, opened 2018-05-17
MAINT: Migrate from Yahoo and Google APIs to Quandl

eigenfoo:
Yahoo has suffered large breaks in the API with no stable replacement. Our current backup, Google, has not been stable since late 2017. Both these errors are upstream, with pandas-datareader, which they are currently trying to fix.

In the meantime, we are using the Quandl API to fetch returns data.

**Note that users must [make a free Quandl account](https://www.quandl.com/) in order to get an API key. This API key must then be set as an environment variable.** In `bash`, this can be done simply by running `export QUANDL_API_KEY='quandl_api_key'`. Without an API key, users may only make up to 50 calls a day, and will not have access to data on certain tickers (most notably SPY).

Clearly, this is an undesirable position to be in: this should be considered a temporary fix until Yahoo or Google Finance transition to more stable APIs, and pandas-datareader starts supporting them again.

Closes #89. Thanks to @jaycode for authoring most of this!

### DIFF
--- a/empyrical/tests/test_stats.py
+++ b/empyrical/tests/test_stats.py
@@ -2,7 +2,6 @@ from __future__ import division
 
 from copy import copy
 from operator import attrgetter
-import random
 from unittest import TestCase, SkipTest
 
 from parameterized import parameterized
@@ -25,6 +24,9 @@ import empyrical.utils as emutils
 DECIMAL_PLACES = 8
 
 
+rand = np.random.RandomState(1337)
+
+
 class BaseTestCase(TestCase):
     def assert_indexes_match(self, result, expected):
         """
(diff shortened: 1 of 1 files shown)


eigenfoo:
It turns out you need a Quandl API key in order to read SPY returns, among other tickers. Also, without an API key, users are limited to 50 calls per day. I've updated the main description with this info.

empyrical/utils.py
```diff
@@ -458,15 +503,15 @@ def default_returns_func(symbol, start=None, end=None):
...
-                                  start='1/1/1970',
+                                  start=start,
                                   end=datetime.now())
-        rets = rets[start:end]
+        rets = rets[rets.index.isin(pd.bdate_range(start, end))]
     else:
-        rets = get_symbol_returns_from_yahoo(symbol, start=start, end=end)
-
+        rets = get_symbol_returns_from_quandl(symbol, start=start, end=end)
+    rets.sort_index(inplace=True)
```
twiecki: the sort should happen up top.

Issue #65, opened 2017-11-16
Remove yahoo finance


twiecki:
Yahoo finance has been discontinued for good. Google is a fallback (albeit a bad one) but there's no reason to display a warning and confuse users: https://github.com/quantopian/pyfolio/issues/478#event-1345024064

eigenfoo:
Would it be correct to say that all that needs rewriting is [this function in `utils.py`](https://github.com/quantopian/empyrical/blob/master/empyrical/utils.py#L389-L397) (and any other docstrings issues here and there)?

twiecki:
@eigenfoo Correct.

twiecki:
It would be great to have an alternative that goes back further in time. We can probably use the yahoo SPY returns for the past, and then forward-fill with the ones from google.

appletonwebb:
There's the IEX exchange API if it helps, and it requires no registration:

https://iextrading.com/developer/docs/#stocks

twiecki:
@jaycode Yes, that seems like a more stable source!

jaycode:
@twiecki Done, please resolve conflicts here: https://github.com/quantopian/empyrical/pull/89

jaycode:
There is also some minimal updates for Pyfolio module for this to work. See here: https://github.com/jaycode/pyfolio/commit/7ef8d27c4926f7cde2abcf6a15d0950450791633

A working example can be seen here: https://github.com/jaycode/pyfolio/blob/master/pyfolio/examples/single_stock_example.ipynb

Issue #64, opened 2017-11-16
Stripping of name will have no effect if cached Fama-French exists


twiecki:
Pyfolio calls https://github.com/quantopian/empyrical/blob/master/empyrical/utils.py#L309 which reads the Fama-French factors from a cached file if it exists. Thus, users upgrading pyfolio and empyrical will get the error that  column `'Mom'` does not exist, as the old name with the white-space is still in the cache.
 
The problem is in this PR: https://github.com/quantopian/empyrical/pull/63

CC @vikram-narayan

vikram-narayan:
if you clear your local cache, this problem will be fixed; do you think it's worth adding a check for this particular case? happy to add one if you think so, just not sure how common the issue is

twiecki:
Yes, but a user will not know what to do and just receive a very odd error.

PR #97, opened 2018-05-18
DEP: Deprecate all data reading functionality via pandas-datareader

eigenfoo:
Closes #96 #89 #65 #64 

Deprecates all data-fetching functionality via `pandas-datareader`, and adds a `simple_returns` helper function to replace that functionality that was implicitly done in the data-fetchers.

To quote the REAME:

As of early 2018, Yahoo Finance has suffered major API breaks with no stable replacement, and the Google Finance API has not been stable since late 2017 [(source)](https://github.com/pydata/pandas-datareader/blob/da18fbd7621d473828d7fa81dfa5e0f9516b6793/README.rst). In recent months it has become a greater and greater strain on the `empyrical` development team to maintain support for fetching data through `pandas-datareader` and other third-party libraries, as these APIs are known to be unstable.

As a result, all `empyrical` support for data reading functionality has been deprecated and will be removed in a future version.

Users should beware that the following functions are now deprecated:

- `empyrical.utils.cache_dir`
- `empyrical.utils.data_path`
- `empyrical.utils.ensure_directory`
- `empyrical.utils.get_utc_timestamp`
- `empyrical.utils._1_bday_ago`
- `empyrical.utils.get_fama_french`
- `empyrical.utils.load_portfolio_risk_factors`
- `empyrical.utils.default_returns_func`
- `empyrical.utils.get_symbol_returns_from_yahoo`

Users should expect regular failures from the following functions, pending patches to the Yahoo or Google Finance API:

- `empyrical.utils.default_returns_func`
- `empyrical.utils.get_symbol_returns_from_yahoo`

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,6 @@ Users should beware that the following functions are now deprecated:
 - `empyrical.utils.cache_dir`
 - `empyrical.utils.data_path`
 - `empyrical.utils.ensure_directory`
-- `empyrical.utils._1_bday_ago`
 - `empyrical.utils.get_fama_french`
 - `empyrical.utils.load_portfolio_risk_factors`
 - `empyrical.utils.default_returns_func`
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ As a result, all `empyrical` support for data reading functionality has been
 deprecated and will be removed in a future version.
 
 Users should beware that the following functions are now deprecated:
+
     - `empyrical.utils.get_fama_french`
     - `empyrical.utils.load_portfolio_risk_factors`
     - `empyrical.utils.default_returns_func`
@@ -84,17 +85,21 @@ Users should beware that the following functions are now deprecated:
 
 Users should expect regular failures from the following functions, pending
 patches to the Yahoo or Google Finance API:
+
     - `empyrical.utils.default_returns_func`
     - `empyrical.utils.get_symbol_returns_from_yahoo`
 
 As an alternative data source, we suggest the following:
+
     1. Migrate your research workflow to the Quantopian Research environment,
        where there is [free and flexible data access to over 57
        datasets](https://www.quantopian.com/data)
     2. Make use of any remaining functional APIs supported by
        `pandas-datareader`. These include:
+
        - [Morningstar](https://pydata.github.io/pandas-datareader/stable/remote_data.html#remote-data-morningstar)
        - [Quandl](https://pydata.github.io/pandas-datareader/stable/remote_data.html#remote-data-quandl)
+
        Please note that you may need to create free accounts with these data
        providers and receive an API key in order to access data. These API keys
        should be set as environment variables, or passed as an argument to
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ patches to the Yahoo or Google Finance API:
 - `empyrical.utils.default_returns_func`
 - `empyrical.utils.get_symbol_returns_from_yahoo`
 
-As an alternative data source, we suggest the following:
+For alternative data sources, we suggest the following:
 
 1. Migrate your research workflow to the Quantopian Research environment,
    where there is [free and flexible data access to over 57
--- a/README.md
+++ b/README.md
@@ -78,6 +78,11 @@ deprecated and will be removed in a future version.
 
 Users should beware that the following functions are now deprecated:
 
+- `empyrical.utils.cache_dir`
+- `empyrical.utils.data_path`
+- `empyrical.utils.ensure_directory`
+- `empyrical.utils.get_utc_timestamp`
+- `empyrical.utils._1_bday_ago`
 - `empyrical.utils.get_fama_french`
 - `empyrical.utils.load_portfolio_risk_factors`
 - `empyrical.utils.default_returns_func`
--- a/README.md
+++ b/README.md
@@ -93,22 +93,6 @@ patches to the Yahoo or Google Finance API:
 - `empyrical.utils.default_returns_func`
 - `empyrical.utils.get_symbol_returns_from_yahoo`
 
-For alternative data sources, we suggest the following:
-
-1. Migrate your research workflow to the Quantopian Research environment,
-   where there is [free and flexible data access to over 57
-   datasets](https://www.quantopian.com/data)
-2. Make use of any remaining functional APIs supported by
-   `pandas-datareader`. These include:
-
-   - [Morningstar](https://pydata.github.io/pandas-datareader/stable/remote_data.html#remote-data-morningstar)
-   - [Quandl](https://pydata.github.io/pandas-datareader/stable/remote_data.html#remote-data-quandl)
-
-   Please note that you may need to create free accounts with these data
-   providers and receive an API key in order to access data. These API keys
-   should be set as environment variables, or passed as an argument to
-   `pandas-datareader`.
-
 ## Contributing
 
 Please contribute using [Github Flow](https://guides.github.com/introduction/flow/). Create a branch, add commits, and [open a pull request](https://github.com/quantopian/empyrical/compare/).
--- a/README.md
+++ b/README.md
@@ -81,7 +81,6 @@ Users should beware that the following functions are now deprecated:
 - `empyrical.utils.cache_dir`
 - `empyrical.utils.data_path`
 - `empyrical.utils.ensure_directory`
-- `empyrical.utils.get_utc_timestamp`
 - `empyrical.utils._1_bday_ago`
 - `empyrical.utils.get_fama_french`
 - `empyrical.utils.load_portfolio_risk_factors`
--- a/README.md
+++ b/README.md
@@ -63,6 +63,43 @@ roll_up_capture(returns, window=60)
 
 Please [open an issue](https://github.com/quantopian/empyrical/issues/new) for support.
 
+### Deprecated: Data Reading via `pandas-datareader`
+
+As of early 2018, Yahoo Finance has suffered major API breaks with no stable
+replacement, and the Google Finance API has not been stable since late 2017
+[(source)](https://github.com/pydata/pandas-datareader/blob/da18fbd7621d473828d7fa81dfa5e0f9516b6793/README.rst).
+In recent months it has become a greater and greater strain on the `empyrical`
+development team to maintain support for fetching data through
+`pandas-datareader` and other third-party libraries, as these APIs are known to
+be unstable.
+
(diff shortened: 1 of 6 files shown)


eigenfoo:
@twiecki @richafrank requesting research and engineer reviews

twiecki:
LGTM. One thing we should add, though, is a function that takes prices and computes correctly indexed returns which one can pass into `pyfolio`.

eigenfoo:
@twiecki done

twiecki:
Thanks, this definitely requires engineering review. CC @richafrank

empyrical/utils.py
```diff
@@ -24,6 +24,16 @@
 import pandas as pd
 from pandas.tseries.offsets import BDay
 from pandas_datareader import data as web
```
eigenfoo: @twiecki @richafrank there is a known import error with `pandas_datareader`: it has been fixed upstream [here](https://github.com/pydata/pandas-datareader/pull/520), but has yet to be released in a new version. See [this StackOverflow thread](https://stackoverflow.com/questions/50394873/import-pandas-datareader-gives-importerror-cannot-import-name-is-list-like).

Wondering what we should do about this. Perhaps wrap it in a try-except, and raise a warning if the import fails? If we're deprecating `pandas-datareader` functionality anyways, it doesn't make sense to keep this line around if it raises an import error.

twiecki:
That sounds like a good solution to me.

eigenfoo:
Done.

empyrical/utils.py
```diff
@@ -205,14 +224,39 @@ def ensure_directory(path):
...
+
+    Returns
+    -------
+    returns : pd.Series or pd.DataFrame
+        Returns of assets in wide-format, with assets as columns,
+        and index coerced to be tz-aware.
+    """
+
+    rets = prices.pct_change().dropna()
+    rets.index = rets.index.tz_localize('UTC')
```
ssanderson: Same note as on your other PR re: not updating values in place here.

empyrical/utils.py
```diff
@@ -205,14 +224,39 @@ def ensure_directory(path):
             raise
 
 
+def compute_returns(prices):
```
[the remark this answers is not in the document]
ssanderson: If we do want this, I would probably call this something like `simple_returns` to distinguish it from log returns. As a general naming note, I'm usually not a fan of including generic verbs like `compute` in function names. All functions compute things, so the `compute_` doesn't really add any useful information for a reader.

empyrical/utils.py
```diff
@@ -205,14 +224,39 @@ def ensure_directory(path):
...
+        and indexed by datetimes.
+
+    Returns
+    -------
+    returns : pd.Series or pd.DataFrame
+        Returns of assets in wide-format, with assets as columns,
+        and index coerced to be tz-aware.
+    """
+
+    rets = prices.pct_change().dropna()
```
ssanderson: Calling `dropna()` and modifying the index here feel philosophically out of place to me for empyrical. 

In my mind at least, the primary purpose of empyrical is to be low-level infrastructure that can be shared betwen Zipline, PyFolio, and Alphalens. Zipline pretty much never wants to call a function that silently discards bad data, or that changes the localization of a datetime index, because both of those behaviors prevent us from noticing bugs, and they incur nontrivial performance overhead.

Calling `dropna()` in particular here seems a bit fraught if we expect DataFrame input, because the behavior there is that pandas will drop an entire row if there are any NaNs in the row. If we expect the user to have missing data, then that seems like not a great behavior (if they pass enough columns, it seems likely that they'll drop a large percentage of their input data). If we don't expect them to have missing data, then we're just incurring extra cost for no reason.

If we want to include a simple returns function in empyrical, I would probably put it in `stats` and write it as something like:

```python
def simple_returns(prices):
    """Compute simple returns from a timeseries of prices.

    <rest of the docstring>
    """
    if isinstance(prices, (pd.DataFrame, pd.Series)):
        return prices.pct_change().iloc[1:]
    else:
        # Assume ndarray
        out = np.diff(prices, axis=0)
        np.divide(out, prices[:-1], out=out)
        return out
```

empyrical/utils.py
```diff
@@ -23,7 +23,23 @@
...
+from .deprecate import deprecated
+
+DATAREADER_DEPRECATION_WARNING = \
+        ("Yahoo and Google Finance have suffered large API breaks with no "
+         "stable replacement. As a result, any data reading functionality "
+         "in empyrical has been deprecated and will be removed in a future "
+         "version."
+         "\n"
+         "Please use empyrical in the Quantopian Research environment, or "
+         "supply your own data. See README.md for more details.")
```
ssanderson: A user reading this error message isn't necessarily going to know what README.md you mean. I'd probably make this a link the github README, and/or add a section in the docs.

empyrical/utils.py
```diff
@@ -23,7 +23,23 @@
...
+    warnings.warn(msg)
+from .deprecate import deprecated
+
+DATAREADER_DEPRECATION_WARNING = \
+        ("Yahoo and Google Finance have suffered large API breaks with no "
+         "stable replacement. As a result, any data reading functionality "
+         "in empyrical has been deprecated and will be removed in a future "
+         "version."
+         "\n"
+         "Please use empyrical in the Quantopian Research environment, or "
```
ssanderson: My inclination would be to drop the reference to Quantopian Research here, especially if we include a reference to it in the README.

empyrical/utils.py
ssanderson: This also needs a test if we decide to keep it.

empyrical/utils.py
```diff
@@ -205,14 +224,39 @@ def ensure_directory(path):
             raise
 
 
+def compute_returns(prices):
+    """
+    Computes correctly-indexed returns from prices.
```
ssanderson: As a reader of this function, I'm not sure I would know what "correctly-indexed" means in this context. Is there something more specific we can say here?

empyrical/utils.py
```diff
@@ -230,10 +274,12 @@ def get_utc_timestamp(dt):
 _1_bday = BDay()
 
 
+@deprecated(msg=DATAREADER_DEPRECATION_WARNING)
```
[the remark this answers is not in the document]
eigenfoo: It's only called in functions that this PR is deprecating, and since this function is for internal use only (i.e. starts with an underscore), I thought it would make sense to deprecate it. Is that not good practice?

empyrical/utils.py
eigenfoo: This will sound bad but I was calling `dropna` because that was the way it was done before, and I assumed there was a reason for that 😦 

I agree that this function belongs in `stats`, though. The question is whether or not we want `empyrical` to tz-localize anything. Philosophically, I feel as if that should be the responsibility of downstream packages like alphalens and pyfolio, since empyrical is only responsible for low-level financial calculations, and is not meant to be user-facing at all. This relates to the input sanitization PR for pyfolio.

@twiecki thoughts?

empyrical/utils.py
eigenfoo: This stops being a problem if we implement the other changes discussed above.

empyrical/utils.py
eigenfoo: This used to mean 'tz-localized', but with the other comments below I don't think we want to do that after all.

empyrical/utils.py
twiecki: Yes, `.iloc[1:]` is a bit clearer on the intent.

twiecki: And yeah, I'd be happy if we just completely did away with the whole tz-aware thing. Since we don't load any data in the future, we can probably just ignore it and the user has to ensure that it matches up.

README.md
```diff
@@ -63,6 +63,52 @@ roll_up_capture(returns, window=60)
...
+
+Users should expect regular failures from the following functions, pending
+patches to the Yahoo or Google Finance API:
+
+- `empyrical.utils.default_returns_func`
+- `empyrical.utils.get_symbol_returns_from_yahoo`
+
+For alternative data sources, we suggest the following:
+
+1. Migrate your research workflow to the Quantopian Research environment,
```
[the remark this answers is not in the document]
TimShawver: Yea I agree...I've been thinking about some similar questions with qgrid since we're going to direct people to research from there.  It feels OK to me in the qgrid case because we're going to be using research to host a demo of qgrid.  In this case it feels a bit weird because we're not directing them to anything empyrical-specific within research, we're just saying here's another place you can get data.

empyrical/utils.py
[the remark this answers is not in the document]
ssanderson: You could, though the particular message is seems a bit confusing for this function. If this is called from deprecated functions, then I'd also be worried about warning spam from this. Since this is a "private" function (in general, functions prefixed with an `_` are private by convention), it doesn't seem all that important to me to deprecate this.

README.md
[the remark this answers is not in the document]
ehebert: I don't know if making the statement less imperative would help it make it less of an advertisement.
i.e. "we suggest" -> "these options are available".

I lean on the side of removing, so that we don't have to worry about semantics.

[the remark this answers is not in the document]
eigenfoo: Okay, removed all suggestions.

empyrical/utils.py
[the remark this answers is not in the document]
eigenfoo: Removed deprecation warning.

eigenfoo:
Still need to write a test for the `simple_returns` function.

eigenfoo:
Not quite sure why tests are failing. The checks fail on my local machine as well, but using `pdb` doesn't reproduce the error.

eigenfoo:
@twiecki fixed unit tests. Ready for another pass, or merging.

README.md
```diff
@@ -63,6 +63,36 @@ roll_up_capture(returns, window=60)
...
+
+As a result, all `empyrical` support for data reading functionality has been
+deprecated and will be removed in a future version.
+
+Users should beware that the following functions are now deprecated:
+
+- `empyrical.utils.cache_dir`
+- `empyrical.utils.data_path`
+- `empyrical.utils.ensure_directory`
+- `empyrical.utils._1_bday_ago`
```
ssanderson: This isn't deprecated anymore.

eigenfoo: Oops

ssanderson:
@eigenfoo I had one more comment on the README. Otherwise this LGTM.

twiecki:
Thanks @eigenfoo!

Issue #114, opened 2019-08-20
Remove Series.strides from codebase (Pandas deprecation)


cgdeboer:
#### Summary
Pandas 0.24 (or 0.23) deprecates the `strides` attribute of `Series` which is used in this repository. https://pandas.pydata.org/pandas-docs/version/0.24/whatsnew/v0.23.0.html?highlight=strides#deprecations

#### Tasks
- [ ] update the utils to handle the deprecation of the `strides` attribute.

twiecki:
Thanks, want to do a PR?

cgdeboer:
@twiecki I may work on this and #113 over the next week or two.

PR #115, opened 2019-08-22
unrequire bottleneck, adjust README, etc.

### LINKED ISSUE #113: Support Python >3.5
Installing fails using Python 3.6 and later as provided in current conda distribution. Bottleneck package dependency seems to fail to install.

cgdeboer:
Closes #113 and #114

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from setuptools import setup
-import os
 import versioneer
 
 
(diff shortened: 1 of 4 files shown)


cgdeboer:
@twiecki I think that should do it for #113 and #114, also added a test runner to make the dev experience of running the tests a bit more readable. No urgency on this.

setup.py
```diff
@@ -61,16 +60,9 @@
     'numpy>=1.9.2',
     'pandas>=0.16.1',
     'scipy>=0.15.1',
-    "pandas-datareader>=0.2",
-    'bottleneck>=1.0.0'
```
twiecki: If we're removing it here we should either list is an optional or add to the readme that people could install bottleneck in addition for more speed.

twiecki:
Thanks @cgdeboer!

PR #121, opened 2020-06-03
Series.as_matrix was depricated then later removed.

atkinson:
Test suite was failing with pandas >= 1.0.

Test suite now passes.

Series.to_numpy() replaces Series.as_matrix().


atkinson:
All the Pandas on the travis file are ancient.

I'll submit an updated travis.

Do you still want to support Python 2?

PR #122, opened 2020-06-03
updated travis 

atkinson:
Travis test configs were all so old that code no longer supported on Pandas 1.0.x would not work on any of them.

Updated to include pandas 1.0.x, python 3.7 numpy 1.18 and scipy 1.4.x.

Kept a python 2.7 config with the last supported versions of pydata (March 2019).

If this is merged, my other pull request should pass.


atkinson:
Actually, this should probably just be in my other pull request. Will close and merge them.

PR #124, opened 2020-06-03
Fix for pandas 1.0

atkinson:
FIXED for Pandas >= 1.0.0

* as_matrix() was deprecated then removed. to_numpy() works in this situation

* updated travis to include pandas 1.0.x, python 3.7 numpy 1.18 and scipy 1.4.x 
  - previously travis only had very old pandas, and this change is not backwards compatible before Pandas 0.23

* kept a python 2.7 build with the last supported versions of pydata.


twiecki:
Thanks for taking this on! Unfortunately due to being tied to zipline we need to maintain compat with their pandas requirements which are older (currently `pandas>=0.18.1,<=0.22`). Can you keep those tests in? Or at least 0.22?

atkinson:
Hey - we could bridge the gap with a try: except: block - try to_numpy() first, then to_matrix() if that fails. 

Shall I do that? Then the older tests should pass as well.

atkinson:
here it is - I left a comment in the code to explain too.

twiecki:
Thanks!

atkinson:
My pleasure. Any chance we could push a release so I can use as a dependency?

twiecki:
@richafrank I tagged a new release but couldn't upload to pypi due to permissions. Could you do that?

PR #140, opened 2024-01-31
Update setup.py

kinskrig:
possible fix for python 3.12


santoshlite:
Hey, it seems like Quantopian is no longer maintained so I decided to bundle empyrical with one of my libraries to keep it working. By the way, the current bug you address has been fixed in my patched version. See more here: https://github.com/santoshlite/EigenLedger/discussions/128

PR #143, opened 2024-07-26
NINF changed to -np.inf

sanchayjain28:
In Numpy NINF is chaned to -np.inf


santoshlite:
Hey, it seems like Quantopian is no longer maintained so I decided to bundle empyrical with one of my libraries to keep it working. See more here: https://github.com/santoshlite/EigenLedger/discussions/128

mozhongzhou:
我也遇到了同样的问题,修改方法一致, np.NINF修改为-np.inf 以适配numpy2.0

